PR #448, opened 2023-11-15
fix: delete nightly release; add docker compose file in release

mooc9988:


mooc9988:
Nightly release is deleted to generate the right change logs in release.

superhx (approved):

PR #472, opened 2023-11-27
feat(issues471): isolate fast / slow read

superhx:
close https://github.com/AutoMQ/automq-for-kafka/issues/471


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1090,13 +1092,41 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+            } catch {
+              case e: Throwable =>
+                val ex = FutureUtil.cause(e)
+                val fastReadFailFast = ex.isInstanceOf[FastReadFailFastException]
+                if (fastReadFailFast) {
+                  slowFetchExecutor.submit(new Runnable {
+                    override def run(): Unit = {
+                      try {
+                        ReadHint.markReadAll()
+                        doFetchingRecords()
```
mooc9988, line 1119: ReadHint.clear()

superhx: The thread is only used for fetch, it's ok not clear

Chillax-0v0 (approved):

PR #504, opened 2023-12-04
fix(core): avoid reading when checking shouldRoll

mooc9988:
In this pull request, to prevent reading the initial batch in ElasticLogSegment#timeWaitedForRoll, the createTimeStamp in ElasticStreamSegmentMeta is utilized to denote the initial timestamp.


core/src/main/scala/kafka/log/streamaspect/ElasticLogSegment.scala
```diff
@@ -317,12 +317,13 @@ class ElasticLogSegment(val _meta: ElasticStreamSegmentMeta,
   }
 
   def timeWaitedForRoll(now: Long, messageTimestamp: Long): Long = {
```
superhx, line 319: If the segmet is empty, we should not roll

PR #544, opened 2023-12-12
fix(issues540): await all partitions close before metadata listener shutdown

superhx:
https://github.com/AutoMQ/automq-for-kafka/issues/540

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2336,9 +2336,6 @@ class ReplicaManager(val config: KafkaConfig,
     removeAllTopicMetrics()
 
     slowFetchExecutors.shutdown()
-
-    awaitAllPartitionShutdown()
-
     info("Shut down completely")
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2780,19 +2769,14 @@ class ReplicaManager(val config: KafkaConfig,
...
+    checkAllPartitionClosed()
+    info("await all partitions closed")
   }
 
   def checkAllPartitionClosed(): Boolean = {
-    val allClosed = allPartitions.isEmpty && closingPartitions.isEmpty
+    val allClosed = allPartitions.isEmpty && closingPartitions.isEmpty && openingPartitions.isEmpty
     if (!allClosed) {
-      info(s"online partition ${allPartitions.keys}, closing partition ${closingPartitions.keys}")
+      info(s"online partition ${allPartitions.keys}, opening partition ${closingPartitions}, closing partition ${closingPartitions.keySet()}")
```
mooc9988, line 2779: opening partition ${openingPartitions}

mooc9988 (approved):

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2760,18 +2753,14 @@ class ReplicaManager(val config: KafkaConfig,
   }
 
   def awaitAllPartitionShutdown(): Unit = {
-    var start = System.currentTimeMillis()
+    val start = System.currentTimeMillis()
     // await 5s partitions transfer to other alive brokers.
```
Chillax-0v0, line 2757: Comments here should be updated too.

PR #590, opened 2023-12-21
fix: fence producing if the first seq is not 0 after creating partition

mooc9988:


core/src/main/scala/kafka/log/ProducerStateManager.scala
```diff
@@ -230,7 +233,15 @@ private[log] class ProducerAppendInfo(val topicPartition: TopicPartition,
     }
   }
 
-  private def checkSequence(producerEpoch: Short, appendFirstSeq: Int, offset: Long): Unit = {
+  // AutoMQ for Kafka inject start
+  private def checkSequence(producerEpoch: Short, appendFirstSeq: Int, offset: Long, logOpenedTimestampOpt: Option[Long]): Unit = {
+    // The partition is newly created while the seq is not 0.
+    if (appendFirstSeq != 0 && updatedEntry.isEmpty && logOpenedTimestampOpt.isDefined && System.currentTimeMillis() - logOpenedTimestampOpt.get < 10000) {
```
superhx, line 239: updatedEntry.isEmpt should put to the first, in most cases, it is possible to return as soon as possible.

superhx (approved):

Issue #586, opened 2023-12-21
Closing partitions cost a lot during shutdown


Chillax-0v0:
server.log:
```
[2023-12-21 02:55:59,988] INFO [Partition test-topic-0000000-5000-687 broker=1] Closing partition (kafka.cluster.Partition)
[2023-12-21 02:56:00,644] INFO [ReplicaManager broker=1] partition test-topic-0000000-5000-687 is closed, trigger leader election (kafka.server.ReplicaManager)
[2023-12-21 02:56:00,645] INFO [Partition test-topic-0000000-5000-1744 broker=1] Closing partition (kafka.cluster.Partition)
[2023-12-21 02:56:00,754] INFO [ReplicaManager broker=1] partition test-topic-0000000-5000-1744 is closed, trigger leader election (kafka.server.ReplicaManager)
[2023-12-21 02:56:00,754] INFO [Partition test-topic-0000000-5000-4651 broker=1] Closing partition (kafka.cluster.Partition)
[2023-12-21 02:56:00,832] INFO [ReplicaManager broker=1] partition test-topic-0000000-5000-4651 is closed, trigger leader election (kafka.server.ReplicaManager)
[2023-12-21 02:56:00,832] INFO [Partition test-topic-0000000-5000-3858 broker=1] Closing partition (kafka.cluster.Partition)
[2023-12-21 02:56:00,973] INFO [ReplicaManager broker=1] partition test-topic-0000000-5000-3858 is closed, trigger leader election (kafka.server.ReplicaManager)
[2023-12-21 02:56:00,973] INFO [Partition test-topic-0000000-5000-2801 broker=1] Closing partition (kafka.cluster.Partition)
[2023-12-21 02:56:01,060] INFO [ReplicaManager broker=1] partition test-topic-0000000-5000-2801 is closed, trigger leader election (kafka.server.ReplicaManager)
[2023-12-21 02:56:01,061] INFO [Partition test-topic-0000000-5000-423 broker=1] Closing partition (kafka.cluster.Partition)
[2023-12-21 02:56:01,161] INFO [ReplicaManager broker=1] partition test-topic-0000000-5000-423 is closed, trigger leader election (kafka.server.ReplicaManager)
[2023-12-21 02:56:01,161] INFO [Partition test-topic-0000000-5000-1480 broker=1] Closing partition (kafka.cluster.Partition)
[...]

PR #591, opened 2023-12-21
perf(ReplicaManager): speed up shutdown

Chillax-0v0:
part of #586


superhx (approved):

Chillax-0v0:
After this PR, it took 45 seconds to close 5000 partitions, down from a timeout (>120s).

PR #651, opened 2024-01-10
perf: return fast when the `fetchOffset` equals to the `confirmOffset`

### LINKED ISSUE #650: When there are a greater number of partitions queried in the fetch, it will allocate more objects
part of #621 

512 topic and 1 partition in each:
cpu:
image
alloc:
image

512 topic and 10 partitions in each:
image
image

Chillax-0v0:
fix #650

### DIFF
--- a/core/src/main/scala/kafka/cluster/Partition.scala
+++ b/core/src/main/scala/kafka/cluster/Partition.scala
@@ -1069,9 +1069,9 @@ class Partition(val topicPartition: TopicPartition,
     // move high watermark based on log confirm offset to prevent ack inflight record.
     leaderLog match {
       case elasticLog: ElasticUnifiedLog =>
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1063,7 +1069,9 @@ class Partition(val topicPartition: TopicPartition,
     // move high watermark based on log confirm offset to prevent ack inflight record.
     leaderLog match {
       case elasticLog: ElasticUnifiedLog =>
-        newHighWatermark = elasticLog.confirmOffset()
+        val cOffset = elasticLog.confirmOffset()
```
superhx, line 1072: confirmOffset

superhx (approved):

PR #671, opened 2024-01-15
feat(core): remove consumer group management for ab consumer

SCNieh:


mooc9988 (approved):

core/src/main/java/kafka/autobalancer/LoadRetriever.java
```diff
@@ -103,30 +105,36 @@ public LoadRetriever(AutoBalancerControllerConfig config, Controller controller,
...
+        this.mainExecutorService.schedule(this::retrieve, 0, TimeUnit.MILLISECONDS);
         logger.info("Started");
     }
 
     public void shutdown() {
         this.shutdown = true;
-        this.executorService.shutdown();
-        retrieveTask.interrupt();
+        this.createTopicExecutorService.shutdown();
+        this.mainExecutorService.shutdown();
```
Chillax-0v0, line 130: `awaitTermination` should be called after `shutdown`

core/src/main/java/kafka/autobalancer/LoadRetriever.java
```diff
@@ -135,10 +143,9 @@ private KafkaConsumer<String, AutoBalancerMetrics> createConsumer(String bootstr
         Properties consumerProps = new Properties();
         consumerProps.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServer);
         consumerProps.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, consumerClientIdPrefix + "-consumer-" + randomToken);
-        consumerProps.setProperty(ConsumerConfig.GROUP_ID_CONFIG, consumerGroupIdPrefix + "-group-" + randomToken);
         consumerProps.setProperty(ConsumerConfig.RETRY_BACKOFF_MS_CONFIG, Long.toString(consumerRetryBackOffMs));
         consumerProps.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest");
-        consumerProps.setProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "true");
+        consumerProps.setProperty(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
```
Chillax-0v0, line 148: Why disable auto commit?

SCNieh: no need to manage offset commits since we always consume from latest offset

PR #719, opened 2024-01-25
feat(metrics): introduce group commit offset metrics

SCNieh:


core/src/main/scala/kafka/coordinator/group/GroupMetadata.scala
```diff
@@ -125,6 +125,7 @@ private[group] case object Empty extends GroupState {
 
 
 private object GroupMetadata extends Logging {
```
superhx, line 127: remove metrics when the coordinator shutdown

superhx (approved):

PR #778, opened 2024-02-05
feat(telemetry): make metrics dashboard compatible with Aliyun

SCNieh:
- remove resource attributes service_namespace and set service_name as cluster_id, thus after translating to prometheus metrics, value of label "job" would be cluster_id
- add additional resource attributes "instance" as node_id to be compatible with Aliyun (Aliyun does not automatically translate service_instance_id to instance)
- remove _total suffix to be compatible with Aliyun

now we have a unified view of labels for most of the OTLP receiver implementation:
image


core/src/main/scala/kafka/log/stream/s3/telemetry/TelemetryManager.java
```diff
@@ -100,12 +100,10 @@ public static boolean isTraceEnable() {
     }
 
     private void init() {
-        String nodeType = getNodeType();
-
         Attributes baseAttributes = Attributes.builder()
-                .put(ResourceAttributes.SERVICE_NAMESPACE, clusterId)
```
daniel-y, line 106: Why remove SERVICE_NAMESPACE?

superhx (approved):

PR #784, opened 2024-02-05
fix(core): release only if it is PooledResource

mooc9988:


fail.yml
```diff
@@ -0,0 +1,4 @@
+2024-02-04--001:
```
superhx, line 1: test file？

superhx (approved):

PR #787, opened 2024-02-05
fix(e2e): wait for more time for broker hard bounce

mooc9988:


superhx:
Check why need recover so long

superhx (approved):

PR #796, opened 2024-02-08
feat(core): verify stream epoch for stream object commit

SCNieh:


core/src/main/scala/kafka/log/streamaspect/MetaStream.java
```diff
@@ -90,6 +90,11 @@ public long streamId() {
         return innerStream.streamId();
     }
 
+    @Override
+    public long streamEpoch() {
```
superhx, line 94: innerStream.streamId

superhx (approved):

PR #850, opened 2024-02-27
perf(ReplicaManager): release unused permits after reads

Chillax-0v0:


Chillax-0v0:
re-PR in https://github.com/AutoMQ/automq-for-kafka/pull/935

PR #900, opened 2024-03-07
feat(config): add config `s3.stream.allocator.policy`

Chillax-0v0:
Note: This PR is not forward compatible, which means, if you want to use direct memory, you need to configure `s3.stream.allocator.policy=POOLED_DIRECT`


Chillax-0v0:
done in https://github.com/AutoMQ/automq-for-kafka/pull/938

PR #982, opened 2024-03-20
fix(s3stream): fix potential concurrent modification on size during t…

SCNieh:


s3stream/src/main/java/com/automq/stream/s3/cache/LogCache.java
```diff
@@ -241,22 +242,29 @@ public void markFree(LogCacheBlock block) {
...
+                if (block.free) {
+                    iter.remove();
+                    remainSize.addAndGet(-block.size());
+                    removed.add(block);
+                } else {
+                    break;
                 }
-                return b.free;
-            });
+            }
```
superhx, line 266: Need to decrease the this.size at the end

superhx (approved):

PR #1005, opened 2024-03-23
feat(core): optimize block cache performance for high-spec node

SCNieh:
1. adjust block cache memory limit and s3 r/w concurrency adaptively
2. add stream set object index to speed up sso lookup
3. increased fetch limiter quota in ReplicaManager
4. increased s3 operator read callback thread pool size
5. increased read ahead agent cache size
6. use seperated thread pool for block cache callback
7. add more comprehensive metrics to trace block cache performance (debug level)

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/image/NodeS3StreamSetObjectMetadataImage.java
+++ b/metadata/src/main/java/org/apache/kafka/image/NodeS3StreamSetObjectMetadataImage.java
@@ -36,6 +36,7 @@ public class NodeS3StreamSetObjectMetadataImage {
     private final int nodeId;
     private final long nodeEpoch;
     private final DeltaMap<Long/*objectId*/, S3StreamSetObject> s3Objects;
+    private final StreamOffsetIndexMap offsetIndexMap = new StreamOffsetIndexMap(2500000);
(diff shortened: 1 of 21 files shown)


Chillax-0v0 (approved):

metadata/src/main/java/org/apache/kafka/image/NodeS3StreamSetObjectMetadataImage.java
```diff
@@ -36,6 +36,7 @@ public class NodeS3StreamSetObjectMetadataImage {
     private final int nodeId;
     private final long nodeEpoch;
     private final DeltaMap<Long/*objectId*/, S3StreamSetObject> s3Objects;
+    private final StreamOffsetIndexMap offsetIndexMap = new StreamOffsetIndexMap(2500000);
```
superhx, line 39: The cache will be deprecated after generating a new image. So the cache may not be available after 1s.

SCNieh: cache will be reconstructed regarding the latest order list, works fine on current scenario

PR #1027, opened 2024-03-26
feat(s3stream): add S3 API compatibility for BOS DeleteObjects

lifepuzzlefun:
BOS baiduyun default DeleteObjects won't return the success deleted object in deleted response. the logic for remove object need this response to mark object success deleted and delete metadata for this object.

So if oss provider not return this fields which will cause object metadata leak and the object status is always in MARK_DESTROYED status and won't be cleaned.

### DIFF
--- a/core/src/main/scala/kafka/log/stream/s3/ConfigUtils.java
+++ b/core/src/main/scala/kafka/log/stream/s3/ConfigUtils.java
@@ -50,7 +50,8 @@ public class ConfigUtils {
                 .networkBaselineBandwidth(s.s3NetworkBaselineBandwidthProp())
                 .refillPeriodMs(s.s3RefillPeriodMsProp())
                 .objectRetentionTimeInSecond(s.s3ObjectDeleteRetentionTimeInSecond())
-                .forcePathStyle(s.s3PathStyle());
+                .forcePathStyle(s.s3PathStyle())
+                .deleteObjectsCompatibilityHandlerClassName(s.s3DeleteObjectsCompatibilityHandlerClassName());
     }
 
 }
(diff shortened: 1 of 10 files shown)


lifepuzzlefun:
server log

> [2024-03-24 00:38:15,917] INFO [ControllerS3Operator]: Delete objects finished, count: 0, cost: 2191439380 (com.automq.stream.s3.operator.DefaultS3Operator) 


metadata snapshot

> >> cat 28
S3Object{objectId=28, objectSize=523341514, objectKey='null', preparedTimeInMs=1711197780365, expiredTimeInMs=1711201380365, committedTimeInMs=1711197785971, markDestroyedTimeInMs=1711200684470, s3ObjectState=MARK_DESTROYED}
>> cat 4110
S3Object{objectId=4110, objectSize=0, objectKey='null', preparedTimeInMs=1711202638476, expiredTimeInMs=1711289038476, committedTimeInMs=0, markDestroyedTimeInMs=1711204854573, s3ObjectState=MARK_DESTROYED}

lifepuzzlefun:
https://cloud.baidu.com/doc/BOS/s/tkc5twspg relate api doc

superhx:
Perhaps we could standardize the delete response handler based on the Error field.

The DeleteObjects quiet mode in AWS S3 follows the same pattern as BOS.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_RequestBody

lifepuzzlefun:
yeah that way can also handle this problem, need more feedback on this ? if change to quiet header is prefered  I will submit another pr to do this

superhx:
> can also handle this proble



> yeah that way can also handle this problem, need more feedback on this ? if change to quiet header is prefered I will submit another pr to do this

Support resolving this issue in a unified way.

lifepuzzlefun:
#1028 submited~

PR #1028, opened 2024-03-26
feat(s3stream): add S3 API compatibility for BOS DeleteObjects

lifepuzzlefun:
BOS baiduyun default DeleteObjects won't return the success deleted object in deleted response. the logic for remove object need this response to mark object success deleted and delete metadata for this object.

So if oss provider not return this fields which will cause object metadata leak and the object status is always in MARK_DESTROYED status and won't be cleaned.

change:
1. change to DeleteObjects quiet mode as default.
2. handle errors in response if no errors think as all delete success.
3. remove those error keys from request deletekeys.


daniel-y:
Great catch, many thanks for this contribution~ @lifepuzzlefun

s3stream/src/main/java/com/automq/stream/s3/operator/DefaultS3Operator.java
```diff
@@ -446,20 +448,63 @@ public CompletableFuture<List<String>> delete(List<String> objectKeys) {
...
+        List<String> ans = null;
+        int errDeleteCount = 0;
+
+        if (!response.hasErrors()) {
+            ans = objectKeys;
+        } else {
+            Set<String> successDeleteKeys = new HashSet<>(objectKeys);
+
+            for (S3Error error : response.errors()) {
+                if (!error.code().equals("NoSuchKey")) {
```
daniel-y, line 490: "NoSuchKey".equals(...)?

s3stream/src/main/java/com/automq/stream/s3/operator/DefaultS3Operator.java
```diff
@@ -446,20 +448,63 @@ public CompletableFuture<List<String>> delete(List<String> objectKeys) {
...
+                    successDeleteKeys.remove(error.key());
```
daniel-y, line 491: The key may not exist even if the error code isn't `NoSuchKey`?

lifepuzzlefun:
hi, guys, I think that simply configuring the deleteObjects API to quiet mode is fragile, as this mode does not explicitly indicate whether a specific Key was successfully deleted. 
If the key in S3Error is unspecified, we still don't know which keys were successfully deleted. 

Therefore, I've made some modifications：

1. When launching the S3Operator, it will automatically check whether to run in quiet mode to deleteObjects. 
2. If we receive a key-does-not-exist error in non-quiet mode, it will be handled correctly. 
3. If the key in S3Error does not exist, this exception will be ignored, and only the 'deleted' marker will be used to determine successful deletions. 
4. If in quiet mode an S3Error's key is "", it can only be assumed that there was an error with the entire deletion request, expecting the Controller to retry deletion of these objects. 

Additionally, I have added some unit tests.

superhx:
> hi, guys, I think that simply configuring the deleteObjects API to quiet mode is fragile, as this mode does not explicitly indicate whether a specific Key was successfully deleted. If the key in S3Error is unspecified, we still don't know which keys were successfully deleted. and check oss doc I believe that aliyun and tecent works as S3API expected.
> 
> Therefore, I've made some modifications：
> 
> 1. When launching the S3Operator, it will automatically check whether to run in quiet mode to deleteObjects.
> 2. If we receive a key-does-not-exist error in non-quiet mode, it will be handled correctly.
> 3. If the key in S3Error does not exist, this exception will be ignored, and only the 'deleted' marker will be used to determine successful deletions.
> 4. If in quiet mode an S3Error's key is "", it can only be assumed that there was an error with the entire deletion request, expecting the Controller to retry deletion of these objects.
> 
> Additionally, I have added some unit tests.

如果测试确认目前主流的云 S3、OSS、COS、BOS 的 deleteObjects API 的 quiet mode 语义保持一致，建议直接默认使用 quiet mode，然后通过错误来判断是否删除成功。我们尽量保持只需要一种判断方式，减少脑力开销。

If tests confirm that the semantics of the quiet mode in the deleteObjects API for the mainstream clouds S3, OSS, COS, and BOS are consistent, it is recommended to use quiet mode by default and then determine the success of the deletion through errors. We strive to maintain only one method of judgment to reduce mental effort.

lifepuzzlefun:
有一个case比较难处理，可能需要看一下怎么弄，

如果quiet模式下返回的Error 字段里的key是空的，可能没办法确认哪些key是删除成功的，因为quiet模式下的处理是 （传进来的全量的key）- (S3Error 返回的error里的key）。但是如果这个S3Error 里面没有key 这个就没法判断了，目前的做法是整个batch 认为删除失败。


主要是感觉只为了百度云把quiet打开有点尴尬，看看怎么处理好弄一点。

superhx:
> 有一个case比较难处理，可能需要看一下怎么弄，
> 
> 如果quiet模式下返回的Error 字段里的key是空的，可能没办法确认哪些key是删除成功的，因为quiet模式下的处理是 （传进来的全量的key）- (S3Error 返回的error里的key）。但是如果这个S3Error 里面没有key 这个就没法判断了，目前的做法是整个batch 认为删除失败。
> 
> 主要是感觉只为了百度云把quiet打开有点尴尬，看看怎么处理好弄一点。
> 
> 如果觉得目前quiet的处理没问题的话，别的代码我就干掉了，统一按照quiet处理了

按照 API 语义来说，如果 response code 200 + 没有 Error Key，是否可以认为都删除成功了，可以研究一下文档 + 实验一下

In terms of API semantics, if the response code is 200 plus there is no Error Key, can we consider that the deletion was successful? It might be worth looking into the documentation and doing some experiments.

superhx:
> https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_ResponseElements 
The operation supports two modes for the response: verbose and quiet. By default, the operation uses verbose mode in which the response includes the result of deletion of each key in your request. In quiet mode the response includes only keys where the delete operation encountered an error. For a successful deletion in a quiet mode, the operation does not return any information about the delete in the response body. 

如果响应为 2xx 且没有 Error，可以认为是全删成功了

If the response code is 2xx and there is no error, it can be considered that everything has been successfully deleted.

lifepuzzlefun:
> > https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_ResponseElements
> > The operation supports two modes for the response: verbose and quiet. By default, the operation uses verbose mode in which the response includes the result of deletion of each key in your request. In quiet mode the response includes only keys where the delete operation encountered an error. For a successful deletion in a quiet mode, the operation does not return any information about the delete in the response body.
> 
> 如果响应为 200 且没有 Error，可以认为是全删成功了
> 
> If the response code is 200 and there is no error, it can be considered that everything has been successfully deleted.

okk, 那如果有Error而且不是KeyNotExist 的先认为整个删除的batch失败吧

superhx:
> 的先认为整个删除的

要不仍旧保持一致的逻辑：不在 Error List 或者 KeyNotExist 都认为成功，然后失败的部分打印一下错误日志

Let's still maintain consistent logic: consider it a success if it's not in the Error List or if KeyNotExist, and then print the error log for the failed part.

s3stream/src/main/java/com/automq/stream/s3/operator/DefaultS3Operator.java
```diff
@@ -442,24 +444,73 @@ public CompletableFuture<Void> delete(String path) {
...
+                hasUnExpectedResponse = true;
+            }
+
+            LOGGER.error("[ControllerS3Operator]: Delete objects for key [{}] error code [{}] message [{}]",
+                    error.key(), error.code(), error.message());
+
+            errDeleteCount++;
+        }
+
+        if (hasUnExpectedResponse) {
```
superhx, line 504: Rely on the semantics of DeleteObjects. Both NoSuchKey and those not on the list are considered to be deleted successfully.

s3stream/src/main/java/com/automq/stream/s3/operator/DefaultS3Operator.java
```diff
@@ -442,24 +444,73 @@ public CompletableFuture<Void> delete(String path) {
...
+                            .record(timerUtil.elapsedAs(TimeUnit.NANOSECONDS));
+                    LOGGER.info("[ControllerS3Operator]: Delete objects failed, count: {}, cost: {}, ex: {}",
+                            objectKeys.size(), timerUtil.elapsedAs(TimeUnit.NANOSECONDS), ex.getMessage());
+                    return Collections.emptyList();
+                });
+    }
+
+    static List<String> handleDeleteObjectsResponse(List<String> objectKeys,
+                                                    TimerUtil timerUtil,
+                                                    DeleteObjectsResponse response) {
```
superhx, line 480: The response code is not 200, did the S3Client return it through thenApply or exceptionally?

If it is returned through thenApply, then it is also necessary to judge the code of the response.

s3stream/src/main/java/com/automq/stream/s3/operator/DefaultS3Operator.java
```diff
@@ -442,24 +444,73 @@ public CompletableFuture<Void> delete(String path) {
...
+                    error.key(), error.code(), error.message());
+
+            errDeleteCount++;
+        }
+
+        if (hasUnExpectedResponse) {
+            successDeleteKeys = Collections.emptySet();
+        }
+
+        LOGGER.info("[ControllerS3Operator]: Delete objects finished, count: {}, errCount: {}, cost: {}",
```
superhx, line 508: If has error, we could print the first N error key and error code

superhx:
> hi, guys, I think that simply configuring the deleteObjects API to quiet mode is fragile, as this mode does not explicitly indicate whether a specific Key was successfully deleted. If the key in S3Error is unspecified, we still don't know which keys were successfully deleted. and check oss doc I believe that aliyun and tecent works as S3API expected.
> 
> Therefore, I've made some modifications：
> 
> 1. When launching the S3Operator, it will automatically check whether to run in quiet mode to deleteObjects.
> 2. If we receive a key-does-not-exist error in non-quiet mode, it will be handled correctly.
> 3. If the key in S3Error does not exist, this exception will be ignored, and only the 'deleted' marker will be used to determine successful deletions.
> 4. If in quiet mode an S3Error's key is "", it can only be assumed that there was an error with the entire deletion request, expecting the Controller to retry deletion of these objects.
> 
> Additionally, I have added some unit tests.

在实际做过测试后，倾向于采用如下方式：
- 在 DefaultS3Operator 启动的时候做探测，看是否支持返回 deleted key，如果支持，则走原有的模式
- 如果不支持则走百度云的模式
- 都为非 quiet 模式

测试结果：
- S3、OSS 如果删除一个不存在的 object，仍旧会在 deleted 列表中返回，没有错误
- OSS DeleteMultipleObjects quiet 模式是什么都不返回，开启感觉有风险 https://www.alibabacloud.com/help/zh/oss/developer-reference/deletemultipleobjects#reference-ydg-25v-wdb

@lifepuzzlefun

lifepuzzlefun:
I believe the biggest risk is that the object storage does not clearly indicate whether the deletion was successful, but the S3Operator returns a success message. In this case, the data is deleted from the metadata, but it is actually still stored in the object storage. I think this kind of leakage is also a significant issue.

So the different here is when the error condition in response won't ensure object to be delete think as the whole delete batch will fail.

superhx (approved):

SCNieh (approved):

superhx (approved):

PR #1035, opened 2024-03-27
fix(core): enhancement AnomalyDetector and LoadRetriever

### LINKED ISSUE #989: [Discuss] Error interrupt LoadRetriever loop
Hi, AutoMQ community, an `Error` will causes the loop to exit, maybe need catch it let the task continue or provide metrics for task state.

Detail see, https://github.com/AutoMQ/automq/blob/baf30f069c58bdb7c88bf9a246cbd45630ea9793/core/src/main/java/kafka/autobalancer/LoadRetriever.java#L431-L435

Wzy19930507:
close #989

* Enhancement AnomalyDetector log throwable and exit loop.
* Enhancement LoadRetriever log throwable and exit loop.

### DIFF
--- a/core/src/main/java/kafka/autobalancer/LoadRetriever.java
+++ b/core/src/main/java/kafka/autobalancer/LoadRetriever.java
@@ -428,9 +428,12 @@ public class LoadRetriever implements BrokerStatusListener {
                 createTopic();
                 this.mainExecutorService.schedule(this::retrieve, 1, TimeUnit.SECONDS);
                 return;
-            } catch (Throwable throwable) {
-                logger.error("Consumer poll throwable", throwable);
+            } catch (Exception e) {
+                logger.error("Consumer poll error", e);
                 this.mainExecutorService.schedule(this::retrieve, 1, TimeUnit.SECONDS);
+                return;
+            } catch (Throwable t) {
+                logger.error("Consumer poll error and exit retrieve loop", t);
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/kafka/autobalancer/LoadRetriever.java
```diff
@@ -428,8 +428,8 @@ public void retrieve() {
                 createTopic();
                 this.mainExecutorService.schedule(this::retrieve, 1, TimeUnit.SECONDS);
                 return;
-            } catch (Exception e) {
```
SCNieh, line 431: We rerun the task on `Exception` is because `Exception` is expected to be recoverable, however for non-recoverable cases (like most of the `Error`), schedule another task is unneccessary becasue the current JVM status is unpredictable, so maybe just add catch `Throwable` after `Exception`, and logging it properly

core/src/main/java/kafka/autobalancer/LoadRetriever.java
```diff
@@ -428,8 +428,8 @@ public void retrieve() {
...
-                logger.error("Consumer poll error", e);
+            } catch (Throwable throwable) {
```
SCNieh, line 431: We rerun the task on `Exception` is because `Exception` is expected to be recoverable, however for non-recoverable cases (like most of the `Error`), schedule another task is unneccessary becasue the current JVM status is unpredictable, so maybe just add catch `Throwable` after `Exception`, and logging it properly

core/src/main/java/kafka/autobalancer/LoadRetriever.java
```diff
@@ -428,8 +428,8 @@ public void retrieve() {
...
+                logger.error("Consumer poll throwable", throwable);
```
SCNieh, line 432: there is no need to change the logging text here, as "throwable" is an adj.

SCNieh (approved):

PR #1039, opened 2024-03-28
[WIP] perf(s3stream): remove lock in streamMetadataManager

lifepuzzlefun:
After check most of kafka origin code. those will try to hold one version of MetadataImage to local variable to finish the request.
the MetadataImage is immutable so the process logic can be threadsafe.

change to volatile fields. when request needed hold as local variable and handle the request.
when the fetch retry the method call will read the current variable to handle the retry logic.


lifepuzzlefun:
relate logic https://github.com/apache/kafka/blob/4cb6806cb8f1cdbf9f47cb6521127fd3f49fa712/core/src/main/scala/kafka/server/metadata/KRaftMetadataCache.scala#L280-L299

PR #1057, opened 2024-03-29
fix(controller): check continuity after redundancy in `commitStreamSetObject`

Chillax-0v0:
The same as #1056


metadata/src/main/java/org/apache/kafka/controller/stream/StreamControlManager.java
```diff
@@ -625,14 +615,16 @@ public ControllerResult<CommitStreamSetObjectResponseData> commitStreamSetObject
                     ControllerResult<Errors> streamObjectCommitResult = this.s3ObjectControlManager.commitObject(streamObject.objectId(),
```
superhx, line 615: Case exists: objectId=-1L, at this point streamObject should be used to determine idempotency.

superhx (approved):

PR #1092, opened 2024-04-07
build(issues1033): switch log appender from DailyRollingFileAppender to RollingFileAppender

warr99:
* Total disk usage: 5,080MB (less than 5GB)
* Details：
image

### DIFF
--- a/tests/kafkatest/services/kafka/templates/log4j.properties
+++ b/tests/kafkatest/services/kafka/templates/log4j.properties
@@ -20,98 +20,86 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
 
 # INFO level appenders
-log4j.appender.kafkaInfoAppender=org.apache.log4j.RollingFileAppender
-log4j.appender.kafkaInfoAppender.MaxFileSize=100MB
-log4j.appender.kafkaInfoAppender.MaxBackupIndex=14
+log4j.appender.kafkaInfoAppender=org.apache.log4j.DailyRollingFileAppender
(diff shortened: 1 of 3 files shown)


zhouxinyu:
Is there any compatible issue if we upgrade the cluster from an old version that already generated some log files? Will the old log files be deleted as expected?

tests/kafkatest/services/kafka/templates/log4j.properties
```diff
@@ -20,86 +20,98 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
 
 # INFO level appenders
-log4j.appender.kafkaInfoAppender=org.apache.log4j.DailyRollingFileAppender
-log4j.appender.kafkaInfoAppender.DatePattern='.'yyyy-MM-dd-HH
+log4j.appender.kafkaInfoAppender=org.apache.log4j.RollingFileAppender
```
superhx, line 23: tests log4j.properties don't need to change

warr99:
The `RollingFileAppender` can only delete files named as `fileName.maxBackupIndex`, and it will not be able to correctly delete log files generated by older versions.

tests/kafkatest/services/kafka/templates/log4j.properties
warr99: Alright, I'll make a change.

zhouxinyu:
> The `RollingFileAppender` can only delete files named as `fileName.maxBackupIndex`, and it will not be able to correctly delete log files generated by older versions.

Got it!

superhx (approved):

Issue #1087, opened 2024-04-03
[Enhancement] BlockCache enhancement


superhx:
TODO: design doc

superhx:
![image](https://github.com/AutoMQ/automq/assets/5719209/dbdefcc2-60ef-4c5a-8d72-0bd4c9662609)

superhx:
[error.log](https://github.com/AutoMQ/automq/files/14967471/error.log)

```
[2024-04-13 05:51:54,186] ERROR [stream-reader-0]Error running task (com.automq.stream.utils.threads.EventLoop)
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.internal.ReferenceCountUpdater.toLiveRealRefCnt(ReferenceCountUpdater.java:83)
	at io.netty.util.internal.ReferenceCountUpdater.release(ReferenceCountUpdater.java:148)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release(AbstractReferenceCountedByteBuf.java:101)
	at io.netty.buffer.AbstractDerivedByteBuf.release0(AbstractDerivedByteBuf.java:98)
	at io.netty.buffer.AbstractDerivedByteBuf.release(AbstractDerivedByteBuf.java:94)
	at com.automq.stream.s3.ObjectReader$DataBlockGroup.release(ObjectReader.java:465)
	at com.automq.stream.s3.cache.blockcache.DataBlock.free(DataBlock.java:62)
	at com.automq.stream.s3.cache.blockcache.DataBlockCache$Cache.evict0(DataBlockCache.java:193)
	at com.automq.stream.utils.threads.EventLoop.run(EventLoop.java:48)
```

```
[2024-04-13 06:49:49,908] INFO [QuorumController id=0] [CommitStreamSetObject]: successfully commit stream set object. streamSetObjectId=37816, nodeId=7, nodeEpoch=1712979070826, compacted objects: [36992, 37761, 35841, 37633, 37762, 35843, 35459, 37763, 37635, 35845, 36998, 36999, 34952, 36363, 37387, 37004, 36878,

[2024-04-13 06:59:53,190] ERROR GetObject for object e0090000/_kafka_ChaosInjectionClusterw/36878 [371469653, 372495902) fail (com.automq.stream.s3.operator.DefaultS3Operator)
[...]

PR #1117, opened 2024-04-13
fix(issues1087): event loop reject task after a task fail 

superhx:
fix https://github.com/AutoMQ/automq/issues/1087#issuecomment-2052725104


Chillax-0v0:
LGTM

PR #1155, opened 2024-04-23
feat(s3stream): change the checksum algorithm to crc32c

ShadowySpirits:
Change the checksum algorithm for the S3 client to reduce CPU usage.

image


superhx (approved):

daniel-y:
Good point! It would be beneficial to include the CPU time for the crc32c algorithm here.

PR #1177, opened 2024-04-25
docs: Fix incorrect redpanda columns in README table

danthegoodman1:
Source: I've used RedPanda _extensively_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ AutoMQ: A Cloud-Native fork of Kafka by separating storage to S3
     <td>In hours/days</td>
     <td>In hours</td>
     <td>In hours<br>(scale-in);<br> In seconds<br>(scale-out)</td>
-    <td>In hours</td>
+    <td>In seconds</td>
     <td>In seconds</td>
   </tr>
   <tr>
@@ -110,7 +110,7 @@ AutoMQ: A Cloud-Native fork of Kafka by separating storage to S3
     <td>In hours/days</td>
     <td>In hours</td>
     <td>In seconds</td>
-    <td>In hours</td>
+    <td>In seconds</td>
     <td>In seconds</td>
   </tr>
   <tr>


daniel-y:
Hi @danthegoodman1, thank you for pointing that out.

In our comparison chart, we've claimed that Redpanda is source-code available, so our comparison is solely with the community edition of Redpanda. It appears that the capability for rapid scaling is dependent on the tiered storage feature, which is a characteristic of the enterprise version.

Even for the enterprise version's tiered storage feature, I still have a question. In order to achieve scaling in seconds, the initial_retention_local_target_ms_default for each partition needs to be set to a smaller value. In cases where the partition size is quite large, it is undeniable that this will result in thousands or even tens of thousands of S3 accesses per second, which is a very expensive cost. Are you aware of this, or have I understood it correctly?

danthegoodman1:
I'm not sure as I haven't monitored that since I don't consider that expensive: At an MSRP of $0.0004/1,000 calls, even 500k calls would be $2. Odds are you're paying more than that per hour for all the instances.

I don't think it's fair to consider the community edition only if you're considering paid-only solutions such as Warpstream or Confluent in the comparison chart. If you use RedPanda cloud, that includes all enterprise functionality (like most managed storage services).

KaimingWan:
@danthegoodman1 Thank you for the information and suggestions you've provided. As you've mentioned, in order to make the content of this table more accurate and objective, and to avoid misunderstandings, we need to make some adjustments to its content. Could you please change the content in the PR to "In hours/In seconds (Enterprise Only)"? With such changes, we should be able to avoid misunderstandings.

daniel-y:
> I'm not sure as I haven't monitored that since I don't consider that expensive: At an MSRP of $0.0004/1,000 calls, even 500k calls would be $2. Odds are you're paying more than that per hour for all the instances.
> 
> I don't think it's fair or consistent to consider the community edition only if you're considering paid-only solutions such as Warpstream or Confluent in the comparison chart. Also if you use RedPanda cloud, IIRC that includes all enterprise functionality (like most managed storage services).

Hi, thanks for the reply, If we need 1000 calls per second for one thousand partitions, 500k calls would be consumed in less than ten minutes. Could you please share more details about your cluster, we may do some accurate calculation work here.

Regarding WarpStream or Confluent, we have labeled them as closed-source software. As Redpanda, @KaimingWan has a proposal to note the commercial ability in the table.

README.md
```diff
@@ -92,7 +92,7 @@ AutoMQ: A Cloud-Native fork of Kafka by separating storage to S3
     <td>In hours/days</td>
     <td>In hours</td>
     <td>In hours<br>(scale-in);<br> In seconds<br>(scale-out)</td>
-    <td>In hours</td>
+    <td>In seconds</td>
```
KaimingWan, line 95: We can use "In hours/In seconds (Enterprise Only)" here. What do you think?

README.md
```diff
@@ -110,7 +110,7 @@ AutoMQ: A Cloud-Native fork of Kafka by separating storage to S3
     <td>In hours/days</td>
     <td>In hours</td>
     <td>In seconds</td>
-    <td>In hours</td>
+    <td>In seconds</td>
```
KaimingWan, line 113: We can use "In hours/In seconds (Enterprise Only)" here. What do you think?

README.md
KaimingWan, line 95: We can use "In hours/In seconds (Enterprise Only)" here. What do you think?

KaimingWan (changes requested):

README.md
danthegoodman1: Real users of any log are likely using enterprise anyway, so best to compare what they are most likely to be using. i.e. these things don't matter to hobbyists who would be using the community edition, so it's most fair to fit the comparison to the target users.

It's unlikely that someone is deciding between the two solutions on these two lines alone.

KaimingWan:
@danthegoodman1 Hi, we have already updated the readme to offer more accurate comparision, so we close the PR now. Thanks for your advice.

Issue #1169, opened 2024-04-24
[Enhancement] Optimize Partition Reassignment logic to avoid NotLeaderOrFollowerException


superhx:
The current Partition Reassignment process, taking the reassignment of P1 from B1 to B2 as an example:
1. Set PartitionRegistration{isr=[B2], replicas=[B2]};
2. B1, upon realizing it is no longer responsible for the partition, begins to shut down;
3. After B1 has completed its partition shutdown, it notifies the Controller to elect a new leader;
4. The Controller sets PartitionRegistration{leader=B2, isr=[B2], replicas=[B2]};
5. B2, upon receiving the notification, starts to open;
6. Once the partition is fully open, it provides messaging capabilities;
Between steps 4 and 6, which lasts about 1.x seconds, Producers/Consumers will receive NotLeaderOrFollowerException and retry their requests.

Expected process:
1. Set PartitionRegistration{isr=[B2], replicas=[B2]};
2. B1, upon realizing it is no longer responsible for the partition, begins to shut down;
3. B2, upon receiving information through Kraft that the streams associated with the partition are closed, begins to open the partition
4. After B2 has opened the partition, it triggers a leader election with the Controller. The Controller sets PartitionRegistration{leader=B2, isr=[B2], replicas=[B2]}

Between steps 1 to 4, clients periodically update Metadata. After step 4, Producers/Consumers will start sending requests to B2

Compatibility: Added automq.version for the evolution of subsequent AutoMQ features/protocols:

Upgrade process:
1. Upgrade all nodes
2. Run `./bin/kafka-features.sh --bootstrap-server localhost:9095 upgrade --feature automq.version=2` to upgrade to V1 version

New cluster: Automatically uses automq.version = LATEST

daniel-y:
For solution 2, what if the partition is going to move to B3 after step 3 before step 4?

superhx:
> For solution 2, what if the partition is going to move to B3 after step 3 before step 4?

If B2 has already completed opening the relevant Stream:
1. After B2 has finished opening, it will proceed to step 4;
2. B2 closes the Partition and the corresponding Streams;
3. B2 an B3 go through steps 1 to 4 again;

If B2 has not yet opened the Stream, but B3 has preemptively opened the Stream:
1. B2 will fail to open the Partition, throwing a StreamFencedException;
2. B3 continues with steps 3 to 4

PR #1180, opened 2024-04-26
feat(issues1169): optimize partition reassignment logic to avoid NotLeaderOrFollowerException

superhx:
close https://github.com/AutoMQ/automq/issues/1169

### DIFF
--- a/core/src/main/scala/kafka/log/streamaspect/DefaultOpenStreamChecker.java
+++ b/core/src/main/scala/kafka/log/streamaspect/DefaultOpenStreamChecker.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2024, AutoMQ CO.,LTD.
+ *
+ * Use of this software is governed by the Business Source License
+ * included in the file BSL.md
+ *
+ * As of the Change Date specified in that file, in accordance with
+ * the Business Source License, use of this software will be governed
+ * by the Apache License, Version 2.0
+ */
+
+package kafka.log.streamaspect;
+
+import com.automq.stream.s3.metadata.StreamState;
+import kafka.server.metadata.KRaftMetadataCache;
+import org.apache.kafka.common.errors.s3.StreamFencedException;
+import org.apache.kafka.image.S3StreamMetadataImage;
+
+public class DefaultOpenStreamChecker implements OpenStreamChecker {
+    private final KRaftMetadataCache metadataCache;
+
+    public DefaultOpenStreamChecker(KRaftMetadataCache metadataCache) {
+        this.metadataCache = metadataCache;
+    }
+
+    @Override
+    public boolean check(long streamId, long epoch) throws StreamFencedException {
+        S3StreamMetadataImage stream = metadataCache.currentImage().streamsMetadata().getStreamMetadata(streamId);
+        if (stream == null) {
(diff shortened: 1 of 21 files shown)


core/src/main/scala/kafka/log/streamaspect/DefaultOpenStreamChecker.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public boolean check(long streamId, long epoch) throws StreamFencedException {
+        S3StreamMetadataImage stream = metadataCache.currentImage().streamsMetadata().getStreamMetadata(streamId);
+        if (stream == null) {
+            throw new StreamFencedException("streamId=" + streamId + " cannot be found, it may be deleted or not created yet");
+        }
+        if (stream.getEpoch() > epoch) {
+            throw new StreamFencedException("streamId=" + streamId + " with epoch=" + epoch + " is fenced by new epoch=" + stream.getEpoch());
```
daniel-y, line 33: Concat the error message is not human readable, could you please consider using `String.format("StreamId=%s with epoch=%s is fenced by......")`

core/src/main/scala/kafka/log/streamaspect/ElasticLog.scala
```diff
@@ -634,6 +636,7 @@ object ElasticLog extends Logging {
                 stream
             } else {
                 val metaStreamId = Unpooled.wrappedBuffer(value.get()).readLong()
+                awaitStreamReadyForOpen(openStreamChecker, metaStreamId, leaderEpoch, logIdent = logIdent)
```
daniel-y, line 639: Is that OK to block the current thread here?

superhx: > Is that OK to block the current thread here?

The Open/Close Partition process is entirely blocked. Opening/closing involves many stages, and the main reason for this approach is to simplify the coding.

To improve the overall concurrency, there is a CachedThreadPool with maxSize=256 in ElasticReplicaManager.

core/src/main/scala/kafka/server/BrokerFeatures.scala
```diff
@@ -75,16 +76,17 @@ object BrokerFeatures extends Logging {
...
+    features.put(MetadataVersion.FEATURE_NAME, new SupportedVersionRange(
+      MetadataVersion.MINIMUM_KRAFT_VERSION.featureLevel(),
+      if (unstableMetadataVersionsEnabled) {
+        MetadataVersion.latestTesting.featureLevel
+      } else {
+        MetadataVersion.latestProduction.featureLevel
+      }
+    ))
+    features.put(AutoMQVersion.FEATURE_NAME, new SupportedVersionRange(AutoMQVersion.V0.featureLevel(), AutoMQVersion.LATEST.featureLevel()))
+    Features.supportedFeatures(features)
```
Chillax-0v0, line 89: Should comment "AutoMQ inject ..." here.

SCNieh (approved):

core/src/main/scala/kafka/server/BrokerFeatures.scala
superhx: > Should comment "AutoMQ inject ..." here.

Fix another PR

Issue #1153, opened 2024-04-23
[Enhancement] Partition Integrity Detection & Correction in Open State


superhx:
### Why is solving this problem impactful?
ElasticReplicaManager controls the opening and closing of partitions based on TopicsDelta, currently facing the following issues:
- Partition state black box: It's impossible to perceive partitions that are in the process of opening or closing, which might lead to scenarios where partitions get stuck either opening or closing without detection (via logs, Metrics);
- No retry mechanism for failed partition openings: During the process of partition opening, underlying exceptions could cause the partition to fail to open. The partition cannot provide service until it is scheduled to another node again;

Expectations:
- Detect partitions that are continuously in the process of opening or closing, and print warn logs and report metrics
- Compare the expected and current state of partitions to be opened, and perform compensation for opening/closing partitions

PR #1238, opened 2024-05-11
feat(issues1153): partition status self-check

superhx:
- part of https://github.com/AutoMQ/automq/issues/1153


core/src/main/scala/kafka/server/streamaspect/ElasticReplicaManager.scala
```diff
@@ -967,6 +972,8 @@ class ElasticReplicaManager(
             }
           }
         } finally {
+          tracker.release()
+          tracker.closed()
```
Chillax-0v0, line 976: `release` is already called in `closed`:
>        public synchronized void closed() {
>            transition(Status.CLOSED);
>            release();
>        }

superhx: Need double release to remove the tracker

Chillax-0v0 (approved):

PR #1259, opened 2024-05-15
feat(issues1244): quick verification of Self-Balancing

### LINKED ISSUE #1244: [Enhancement] Send a message to the specified brokers
### Who is this for and what problem do they have today?
* The current kafka-producer-perf-test.sh cannot send messages to partitions on the specified broker.

### Why is solving this problem impactful?
* To enable users to quickly experience the partition self-balancing triggered by partition hotspots through AutoMQ, we plan to add a new parameter `--broker <broker1,broker2,...>` to `kafka-producer-perf-test.sh`. Messages will only be sent to the specified brokers, thereby creating partition hotspots and triggering partition self-balancing.

### Additional notes
* org.apache.kafka.tools.ProducerPerformance
* bin/kafka-producer-perf-test.sh

warr99:
close #1244

### DIFF
--- a/tools/src/main/java/org/apache/kafka/tools/ProducerPerformance.java
+++ b/tools/src/main/java/org/apache/kafka/tools/ProducerPerformance.java
@@ -26,14 +26,24 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Arrays;
 import java.util.SplittableRandom;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
 
+import net.sourceforge.argparse4j.impl.Arguments;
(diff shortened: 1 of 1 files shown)


tools/src/main/java/org/apache/kafka/tools/ProducerPerformance.java
```diff
@@ -486,4 +536,33 @@ public void onCompletion(RecordMetadata metadata, Exception exception) {
...
+                throw new IllegalArgumentException("The list of partitions in brokers cannot be null or empty.");
+            }
+
+            this.partitionsInBrokers = Arrays.stream(partitionsInBrokersStr.split(","))
+                .map(Integer::parseInt)
+                .collect(Collectors.toList());
+        }
+
+        @Override
+        public int partition(String topic, Object key, byte[] keyBytes, Object value, byte[] valueBytes, Cluster cluster) {
```
daniel-y, line 555: You seem to have enough information from the `cluster` parameter to determine the unique node ID to send all the data.

So, consider moving all the partition selection logic to this partitioner, you even don't need to set any key for records.

PR #1293, opened 2024-05-23
feat(shell): add s3 log appender

ShadowySpirits:


core/src/main/scala/kafka/Kafka.scala
```diff
@@ -108,22 +110,38 @@ object Kafka extends Logging {
 
   private def buildServer(props: Properties): Server = {
     val config = KafkaConfig.fromProps(props, doLog = false)
+    AutoMQApplication.registerSingleton(classOf[S3LogConfig], new S3LogConfig {
+
+      override def isDebugEnabled: Boolean = config.isDebugEnabled
```
superhx, line 115: log bucket is ops bucket

superhx (approved):

PR #1294, opened 2024-05-23
feat(shell): add s3 metrics exporter

ShadowySpirits:


core/src/main/scala/kafka/log/stream/s3/telemetry/TelemetryManager.java
```diff
@@ -266,6 +271,37 @@ private SdkMeterProvider getMetricsProvider(Resource resource) {
         return sdkMeterProviderBuilder.build();
     }
 
+    private void initS3Exporter(SdkMeterProviderBuilder sdkMeterProviderBuilder, KafkaConfig kafkaConfig) {
```
superhx, line 274: TODO: auto init s3 metrics exporter when ops bucket specified

ShadowySpirits: I will fix it in next pr.

superhx (approved):

PR #1308, opened 2024-05-27
feat: close valid streams first when encountering an invalid stream during recovery

Chillax-0v0:


s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
```diff
@@ -163,13 +170,75 @@ static LogCache.LogCacheBlock recoverContinuousRecords(Iterator<WriteAheadLog.Re
...
+            logger.info("drop discontinuous records, records={}", queue);
+            queue.forEach(StreamRecordBatch::release);
+        });
+
+        if (logEndOffset >= 0L) {
+            cacheBlock.confirmOffset(logEndOffset);
+        }
+
+        RecoverResult result = new RecoverResult();
+        cacheBlock.records().forEach((streamId, records) -> {
```
superhx, line 194: Only add the stream first exception

superhx (approved):

PR #1344, opened 2024-06-03
fix(s3stream): refine network rate limiter for better precision

SCNieh:


s3stream/src/main/java/com/automq/stream/s3/network/AsyncNetworkBandwidthLimiter.java
```diff
@@ -221,8 +243,13 @@ public int compareTo(BucketItem o) {
             return Long.compare(strategy.priority(), o.strategy.priority());
```
superhx, line 243: Add timestamp compare for the same priority

superhx (approved):

PR #1377, opened 2024-06-06
fix(core): add tolerance delay when append with non-zero seq number with empty producer snapshot

SCNieh:


core/src/main/scala/kafka/log/streamaspect/ElasticProducerStateManager.java
```diff
@@ -53,6 +55,7 @@ public ElasticProducerStateManager(
         // ProducerStateManager will loadSnapshots in constructor, but at that moment, snapshotsMap is null.
         // So we need to call loadSnapshots again to load snapshots from snapshotsMap
         this.snapshots = loadSnapshots();
+        this.createTimestamp = time.milliseconds();
```
ShadowySpirits, line 58: It's better to start the timer when receiving a request from a new producer with a sequence ID greater than 0.

superhx (approved):

PR #1411, opened 2024-06-13
perf(s3stream): reduce lock granularity in `StreamMetadataManager`

lifepuzzlefun:
1. use volatile field to store MetadataImage avoid acquire lock when kraft metadata update
2. when access metadata store MetadataImage to local field without hold locks.
3. for the index structure in MetadataImage use lock to trigger index build and after build success the access is volatile access.
4. avoid get list.size() in for iter.

##### test
1. add concurrent access metadata test to check the access is right.

### DIFF
--- a/core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
+++ b/core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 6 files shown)


core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
```diff
@@ -222,4 +237,28 @@ public GetObjectsTask() {
         }
     }
 
+    private static class S3StreamImagePair {
```
superhx, line 240: Why do we need S3StreamImagePair? Images themselves are immutable.

core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
```diff
@@ -74,118 +65,142 @@ public String name() {
...
+                                        "and search in metadataCache success with result: {}",
+                                streamId, startOffset, endOffset, limit, rst);
+                    }
+
+                    return inRangeObjects;
+                });
     }
 
-    private synchronized CompletableFuture<InRangeObjects> fetch0(long streamId, long startOffset, long endOffset, int limit) {
+    private CompletableFuture<Map.Entry<InRangeObjects, S3StreamImagePair>> fetch0(long streamId, long startOffset, long endOffset, int limit) {
```
superhx, line 128: We can move the logic of filling S3Object information from #fetch to #fetch0, so we don't need to return image.

core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
lifepuzzlefun: yes those were immutable. we need atomic replace for the two stream relate metadata Images. 
this package can volatile update and read store in local variable without locks (for multi fields)

core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
lifepuzzlefun: I think the image should be in the same version when get metadata. this pr won't block image update and image retrieve logic. so when the metadata update the images are not the same version which will cause metadata inconsistency.

lifepuzzlefun: yeah i understand what you said. i will check it out

lifepuzzlefun: updated

lifepuzzlefun:
if more unit test need to be added please let me know : - )

superhx (approved):

PR #1441, opened 2024-06-18
feat(s3stream): add s3 delete implementation for object storage 

warr99:


s3stream/src/main/java/com/automq/stream/s3/operator/AwsObjectStorage.java
```diff
@@ -212,13 +253,117 @@ void doClose() {
         }
     }
 
+    static void handleDeleteObjectsResponse(DeleteObjectsResponse response, boolean deleteObjectsReturnSuccessKeys) throws Exception {
+        int errDeleteCount = 0;
+        ArrayList<String> failedKeys = new ArrayList<>();
```
superhx, line 258: Why has failed keys, the response is still success

superhx (approved):

PR #1447, opened 2024-06-20
perf(s3stream): Avoid non-reuse TimerUtil object creation on write path

lifepuzzlefun:


lifepuzzlefun:
maybe a IOTraceContext pass through the write path is much better, but for current code one write request will create a lot TimerUtil, store to primitive may have better performance.

SCNieh (approved):

PR #1469, opened 2024-06-26
fix(s3stream): ensure wal header shutdown type UNGRACEFULLY when startup

lifepuzzlefun:
current code if wal sliding window not expand 
the wal header won't flush to UNGRACEFULLY


Chillax-0v0 (approved):
LGTM

PR #1474, opened 2024-06-27
fix(metadata): fix mark destoryed object queue won't be cleanup in fo…

superhx:
The same as https://github.com/AutoMQ/automq/pull/1473


daniel-y (approved):
LGTM~

PR #1481, opened 2024-06-28
perf(s3stream): Reduce AsyncSemaphore lock contention

lifepuzzlefun:


s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/AsyncSemaphore.java
```diff
@@ -37,34 +38,37 @@ public AsyncSemaphore(long permits) {
...
-    public synchronized long permits() {
-        return permits;
+    public long permits() {
+        return permits.get();
     }
 
-    synchronized void release(long requiredPermits) {
-        permits += requiredPermits;
-        if (permits > 0) {
+    void release(long requiredPermits) {
```
superhx, line 70: Acquire and release may cause concurrency issues:
1. Thread1: acquire long current = permits.get(), finds it is less than 0
2. Thread2: release permits.addAndGet(requiredPermits) > 0, tries to retrieve a task from the tasks queue, but it is still empty at this point
3. Thread1: places the task into tasks, but no one is awakened at this moment

AsyncSemaphore is only used during cold reading, with low concurrency, temporarily not a performance bottleneck point, can be optimized later.

lifepuzzlefun: agree. we can do this later : - )

PR #1542, opened 2024-07-10
feat(metadata): accelerate stream range search with async cache

SCNieh:


SCNieh:
Wait until this becomes a problem

PR #1561, opened 2024-07-12
feat(s3stream): add random jitter delay time for StreamObjectsCompaction

lifepuzzlefun:


lifepuzzlefun:
in our environment all broker start almost in the same time.
the same as stream object compaction. add jitter to avoid them run at the same time
image

superhx (approved):

PR #1518, opened 2024-07-08
fix(s3stream): fix delete composite object exceed S3 API limit

lifepuzzlefun:

PR #1567, opened 2024-07-15
feat(s3stream): batch delete composite object to reduce delete api call

lifepuzzlefun:
1. if 100 composite object each with 4 linked object. try to get all linked object information first and batch in 1000 keys. then one delete api call can delete them. (compare with 200 call previous)

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/stream/S3ObjectControlManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/stream/S3ObjectControlManager.java
@@ -17,7 +17,6 @@
 
 package org.apache.kafka.controller.stream;
 
-import com.automq.stream.s3.CompositeObject;
 import com.automq.stream.s3.Config;
 import com.automq.stream.s3.ObjectReader;
 import com.automq.stream.s3.compact.CompactOperations;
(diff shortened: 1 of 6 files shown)


superhx:
> 1. if 100 composite object each with 4 linked object. try to get all linked object information first and batch in 1000 keys. then one delete api call can delete them. (compare with 200 call previous)

- The Delete API is free. 
- Can it be placed inside AbstractObjectStorage for optimization?

lifepuzzlefun:
> > 1. if 100 composite object each with 4 linked object. try to get all linked object information first and batch in 1000 keys. then one delete api call can delete them. (compare with 200 call previous)
> 
> * The Delete API is free.
> * Can it be placed inside AbstractObjectStorage for optimization?

got it another problem is that when delete a large topic. then a lot of object will be delete (100w+) . the object storage delete api have a call limit which will make the whole process very long

superhx:
> the

ObjectStorage still keeps List<ObjectPath> as a parameter, and then AbstractObjectStorage internally can batch/split the delete list.

lifepuzzlefun:
> > the
> 
> ObjectStorage still keeps List as a parameter, and then AbstractObjectStorage internally can batch/split the delete list.

this logic is introduced by https://github.com/AutoMQ/automq/pull/1518

----
let me explain this PR in more detail.

the ObjectStorage already have batch split logic internally, and only get List<ObjectPath> input

for 100 CompositeObject delete. each object will get its linkedObject first.

and the List<ObjectPath> = 100 CompositeObject + 100 * LinkedObjectPerCompositeObject （which may be 1000 for each）

the objectcleaner collect the first finished get linked object and collect them in batch then submit to objectStorage layer to delete.

and follow all the DeleteComposite CompletableFuture to get the result.

----
the ObjectCleaner do the batch before call the ObjectStorage is because if the 1.5d topic delete then if collect all composite and linked object the in-memory object list may contains 100w+ item which will be make memory pressure to controller

lifepuzzlefun:
new PR https://github.com/AutoMQ/automq/pull/1572

PR #1572, opened 2024-07-16
feat(s3stream): batch delete in AbstractObjectStorage

lifepuzzlefun:
1. add deleteOption to disable batch logic. default delete request will be batched.
2. DeleteObjectsAccumulator will handle the time-based and size-based logic to submit deleteObjects request when condition meet.


s3stream/src/main/java/com/automq/stream/s3/operator/DeleteObjectsAccumulator.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class DeleteObjectsAccumulator implements Runnable {
```
superhx, line 33: Is it possible to indirectly accumulate batches by limiting the number of Inflight Deletes? This way, even in cases where Deletes are relatively few, the latency of Deletes won't be compromised.

superhx: For example, restrict the Delete concurrency to 10. If the Inflight does not exceed 10, it is freely released. If it exceeds 10, it queues up inside, waiting for the release in front before being taken out of the queue.

lifepuzzlefun: current code support config no batch DeleteOptions to do this.
for the frontend request user can use this to skip time batch logic.

superhx: In theory, higher-level code should not be aware of whether it needs to Batch.

superhx: Because the underlying concurrency of S3 is also limited (default 50), in fact, for latency-sensitive APIs, limiting the number of inflights and accumulating batches through inflight restrictions is also better.

lifepuzzlefun: got it. I'll change the code.

s3stream/src/test/java/com/automq/stream/s3/compact/CompactionManagerTest.java
```diff
@@ -563,7 +566,7 @@ public void testCompactionShutdown() throws Throwable {
 
         objectManager = Mockito.spy(MemoryMetadataManager.class);
         objectStorage = Mockito.spy(MemoryObjectStorage.class);
-        List<Pair<InvocationOnMock, CompletableFuture<ByteBuf>>> invocations = new ArrayList<>();
+        final BlockingQueue<Pair<InvocationOnMock, CompletableFuture<ByteBuf>>> invocations = new LinkedBlockingQueue<>();
```
lifepuzzlefun, line 569: this fix is for flaky unit test of testCompactionShutdown the origin invocations is not thread safe.

superhx (approved):

s3stream/src/main/java/com/automq/stream/s3/operator/DeleteObjectsAccumulator.java
```diff
@@ -0,0 +1,158 @@
...
+    static final Logger LOGGER = LoggerFactory.getLogger(DeleteObjectsAccumulator.class);
+    public static final int DEFAULT_DELETE_OBJECTS_MAX_BATCH_SIZE = 1000;
+    public static final int DEFAULT_DELETE_OBJECTS_MAX_CONCURRENT_REQUEST_NUMBER = 100;
+    private final ExecutorService batchDeleteCheckExecutor = Threads.newFixedThreadPoolWithMonitor(1,
```
superhx, line 38: The accumulator could be optimized to without extra thread

lifepuzzlefun: got it I'll try to optimize this.

Issue #1446, opened 2024-06-20
[Enhancement] Add quota for low-priority controller mutations


ShadowySpirits:
### Who is this for and what problem do they have today?

A large number of compaction or delete object requests may overload the controller event loop.
We should add a mechanism to protect the controller by stopping low-priority requests and allowing it to prioritize high-priority requests, such as broker heartbeat.

### Why is solving this problem impactful?

The high-priority request can not fail; otherwise, the cluster may refuse to serve.

DaniilRoman:
Hello 👋 
This issue looks interesting for me and I'd like to take it over, but I'd love to ask few questions upfront 

1. Do I understand it right that the idea, simply saying, is substituting `LinkedBlockingQueue` in `ControllerEventManager` to `PriorityBlockingQueue` ?
2. If I understand this correctly, I see that in [Kafka](https://github.com/apache/kafka/blob/4d3e366bc2487ca0c2321e6e6d7786fb6ed0efa5/core/src/main/scala/kafka/controller/ControllerEventManager.scala#L83) upstream they still use `LinkedBlockingQueue`. Might it be a case that they've already considered this idea and rejected it for any reasons? 
3. Do you have already any statistics about different events and their possibilities to block the event loop? It might be helpful for identifying the types of events that should be with the low priority.
4. Is the any advice of handling the edge case where the low priority might stuck in a queue for a too long? Should we change the low priority for such events to the high one after the specific timeout?

ShadowySpirits:
> Hello 👋 This issue looks interesting for me and I'd like to take it over, but I'd love to ask few questions upfront
> 
> 1. Do I understand it right that the idea, simply saying, is substituting `LinkedBlockingQueue` in `ControllerEventManager` to `PriorityBlockingQueue` ?
> 2. If I understand this correctly, I see that in [Kafka](https://github.com/apache/kafka/blob/4d3e366bc2487ca0c2321e6e6d7786fb6ed0efa5/core/src/main/scala/kafka/controller/ControllerEventManager.scala#L83) upstream they still use `LinkedBlockingQueue`. Might it be a case that they've already considered this idea and rejected it for any reasons?
> 3. Do you have already any statistics about different events and their possibilities to block the event loop? It might be helpful for identifying the types of events that should be with the low priority.
> 
> Looking forward for your answer 🙏

[...]

PR #1576, opened 2024-07-16
feat(metadata): adding ControllerTimeoutDecorator to prevent overloading of the main Controller

DaniilRoman:
PR solves the issue https://github.com/AutoMQ/automq/issues/1446
Added a unit test to test a new functionality 


superhx:
Kafka assumes that requests will be processed sequentially. Changing the queue to a PriorityQueue may introduce unknown risks.

https://github.com/AutoMQ/automq/issues/1446 The core purpose is to solve how to handle scenarios where the Controller (EventLoop / CPU / GC) is overloaded, ensuring that the Controller does not mistakenly consider all Brokers as Fenced, triggering massive partition migrations.

metadata/src/main/java/org/apache/kafka/controller/ControllerTimeoutDecorator.java
```diff
@@ -0,0 +1,403 @@
...
+ */
+public class ControllerTimeoutDecorator implements Controller {
+
+    private final Controller controller;
+
+    private final int defaultTimeoutInMs;
+
+    public ControllerTimeoutDecorator(Controller controller) {
+        this.controller = controller;
+        this.defaultTimeoutInMs = 3000;
```
DaniilRoman, line 106: I assume you have the necessary statistic data to be able to correct the default timeout

metadata/src/main/java/org/apache/kafka/controller/ControllerTimeoutDecorator.java
```diff
@@ -0,0 +1,403 @@
...
+    public CompletableFuture<Integer> getNextNodeId(
+            ControllerRequestContext context,
+            GetNextNodeIdRequestData request
+    ) {
+        return controller.getNextNodeId(context, request);
+    }
+
+    @Override
+    public CompletableFuture<Void> checkS3ObjectsLifecycle(ControllerRequestContext context) {
+        return controller.checkS3ObjectsLifecycle(context).orTimeout(defaultTimeoutInMs, TimeUnit.MILLISECONDS);
```
DaniilRoman, line 287: Added timeout only for 2 new methods and that are running by schedule

core/src/main/scala/kafka/server/ControllerServer.scala
```diff
@@ -581,9 +582,6 @@ class ControllerServer(
   }
 
   // AutoMQ for Kafka inject start
-  protected def quorumControllerExtension(quorumController: QuorumController): QuorumControllerExtension = {
-    QuorumControllerExtension.NOOP
-  }
```
DaniilRoman, line 586: It wasn't used anywhere

core/src/main/scala/kafka/server/ControllerServer.scala
```diff
@@ -117,7 +116,7 @@ class ControllerServer(
   var createTopicPolicy: Option[CreateTopicPolicy] = None
   var alterConfigPolicy: Option[AlterConfigPolicy] = None
   @volatile var quorumControllerMetrics: QuorumControllerMetrics = _
-  var controller: QuorumController = _
+  var controller: Controller = _
```
DaniilRoman, line 119: Initially there was Controller as well

DaniilRoman:
Thank you @superhx for the clarifications 🙏  Initially I was looking at the wrong direction. I found [that docs](https://docs.automq.com/automq/what-is-automq/compatibility-with-apache-kafka#apache-kafka-compatibility-overview) where was said that AutoMQ doesn't even use old ZK mode and instead AutoMQ always runs in KRaft mode

metadata/src/main/java/org/apache/kafka/controller/ControllerTimeoutDecorator.java
superhx: The timeout only complete the CompletableFuture with TimeoutException, the `#checkS3ObjectsLifecycle` is still running in the EventLoop.

superhx:
Do you mind if I solve this issue in this way? https://github.com/AutoMQ/automq/issues/1446

superhx:
https://github.com/AutoMQ/automq/issues/1446 resolved

PR #1577, opened 2024-07-17
feat(controller): limit max checkS3ObjectsLifecycle run time

lifepuzzlefun:


lifepuzzlefun:
In our previous test version the code create a lot of preparedObjects. the check run for very long time cause all broker fenced. also this won't happen too much time in normal. I think we can limit the max runtime avoid harm the eventloop.

```shell
Controller took 12415025 µs to process write event: checkS3ObjectsLifecycle
```

superhx:
> In our previous test version the code create a lot of preparedObjects. (25000w+) the check run for very long time cause all broker fenced. also this won't happen too much time in normal. I think we can limit the max runtime avoid harm the eventloop.
> 
> ```shell
> Controller took 12415025 µs to process write event: checkS3ObjectsLifecycle
> ```

We have also noticed this issue and in the next few days, it will be addressed through a more elegant strategy to avoid the problem of check_lifecycle taking too long.

- Use HashTimeWheel to detect prepare object ttl
- Use swap operation to clean up mark destroyed object https://github.com/AutoMQ/automq/pull/1473

lifepuzzlefun:
cool! look forward to the following feature

PR #1584, opened 2024-07-17
feat(s3stream): support config MAJOR_V1 skip compact small object

lifepuzzlefun:
1. support MAJOR_V1 compaction skip object which size is below the MINOR_V1_COMPACTION_SIZE.
2. default skip logic is disabled.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/S3StreamClient.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3StreamClient.java
@@ -60,7 +60,6 @@ public class S3StreamClient implements StreamClient {
     private static final long COMPACTION_COOLDOWN_AFTER_OPEN_STREAM = Systems.getEnvLong("AUTOMQ_STREAM_COMPACTION_COOLDOWN_AFTER_OPEN_STREAM", TimeUnit.MINUTES.toMillis(1));
     private static final long MINOR_V1_COMPACTION_INTERVAL = Systems.getEnvLong("AUTOMQ_STREAM_COMPACTION_MINOR_V1_INTERVAL", TimeUnit.MINUTES.toMillis(10));
     private static final long MAJOR_V1_COMPACTION_INTERVAL = Systems.getEnvLong("AUTOMQ_STREAM_COMPACTION_MAJOR_V1_INTERVAL", TimeUnit.MINUTES.toMillis(60));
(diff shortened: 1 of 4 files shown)


lifepuzzlefun:
why this pr？
I found that in single bucket the object is very large 1230517 object for 1200 s3Object recored in metadata.

82G total data for 1230517 object in one bucket = 71.58KB / object

I think the major_v1 may compose both the small object and the large object together.

for MINIO the object storage metadata may prefer more bigger object with less total object number.

s3stream/src/main/java/com/automq/stream/s3/compact/StreamObjectCompactor.java
```diff
@@ -116,6 +127,28 @@ private void handleCompactException(CompactionType compactionType, Throwable e)
         }
     }
 
+    protected static Predicate<S3ObjectMetadata> getObjectFilter(CompactionType compactionType, long minMajorV1CompactionSize) {
+        boolean includeCompositeObject = CLEANUP_V1.equals(compactionType) || MAJOR_V1.equals(compactionType);
+
+        return object -> {
+            ObjectAttributes.Type objectType = ObjectAttributes.from(object.attributes()).type();
+            if (!includeCompositeObject && objectType == Composite) {
+                return true;
```
superhx, line 136: return true means include, the if check should be revert

s3stream/src/main/java/com/automq/stream/s3/S3StreamClient.java
```diff
@@ -366,8 +369,15 @@ private void compactV1(CompactionHint hint, long now) {
...
         public void compact(StreamObjectCompactor.CompactionType compactionType) {
-            StreamObjectCompactor task = StreamObjectCompactor.builder().objectManager(objectManager).stream(this)
-                .objectStorage(objectStorage).maxStreamObjectSize(config.streamObjectCompactionMaxSizeBytes()).build();
+            StreamObjectCompactor task = StreamObjectCompactor.builder()
+                .objectManager(objectManager)
+                .stream(this)
+                .objectStorage(objectStorage)
+                .maxStreamObjectSize(config.streamObjectCompactionMaxSizeBytes())
+                .minorV1CompactionThreshold(MINOR_V1_COMPACTION_SIZE)
+                .majorV1CompactionSkipSmallObject(MAJOR_V1_COMPACTION_SKIP_SMALL_OBJECT)
```
superhx, line 378: We can skip the small object logic for objectsCount < MAJOR_V1_COMPACTION_MAX_OBJECT_THRESHOLD instead of by a switch.

superhx (approved):

PR #1596, opened 2024-07-19
test(s3stream): add waiting for InflightReadaheadCf

llzcx:


s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
```diff
@@ -221,13 +221,23 @@ void read0(ReadContext ctx, final long startOffset, final long endOffset, final
...
     public CompletableFuture<Void> getAfterReadTryReadaheadCf() {
         return afterReadTryReadaheadCf;
     }
 
+    /**
+     * inflightReadaheadCf is empty, the task has been completed
+     * inflightReadaheadCf is not empty, the task may be completed
+     * @return  readahead.inflightReadaheadCf
+     */
+    public CompletableFuture<Void> getReadaheadInflightReadaheadCf() {
```
superhx, line 237: The method expose to test should remove public and mark only for test

superhx (approved):

PR #1599, opened 2024-07-19
feat(metadata): use lru cache to control the overall size of node sparse index cache

SCNieh:


metadata/src/main/java/org/apache/kafka/image/S3StreamsMetadataImage.java
```diff
@@ -289,6 +289,21 @@ void fillObjects(
         }
     }
 
+    private void completeWithSanityCheck(GetObjectsContext ctx, List<S3ObjectMetadata> objects) {
```
superhx, line 292: The ctx fields will be changed after #getObjects

s3stream/src/main/java/com/automq/stream/s3/cache/AsyncObjectLRUCache.java
```diff
@@ -0,0 +1,60 @@
...
+        int totalSize = totalSize();
+        while (totalSize > maxSize) {
+            Optional.ofNullable(pop()).ifPresent(entry -> {
+                try {
+                    entry.getValue().close();
+                } catch (Exception e) {
+                    // ignore
+                }
+            });
+            totalSize = totalSize();
```
superhx, line 44: The totalSize could be increase when the value.size() complete

superhx (approved):

PR #1625, opened 2024-07-23
feat(wal): wal start optimization

CLFutureX:
Background:

Currently, the startup process of WAL (Write-Ahead Logging) lacks adequate concurrency safeguards, and improper external invocations can lead to concurrent execution scenarios.

Furthermore, there isn't a well-formed closed loop between the start and shutdown operations: in extreme cases, if the start method and shutdown are executed in an overlapping manner, it can result in a failed shutdown process.

Optimization:

Adjust the start logic to ensure concurrency safety.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
@@ -305,6 +305,7 @@ public class BlockWALService implements WriteAheadLog {
             } finally {
                 if (!success) {
                     started.set(false);
+                    LOGGER.warn("block WAL service started fail");
                 }
             }
         }
--- a/s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
@@ -273,31 +273,42 @@ public class BlockWALService implements WriteAheadLog {
 
     @Override
     public WriteAheadLog start() throws IOException {
-        if (started.get()) {
+        boolean state = started.get();
+        if (state) {
             LOGGER.warn("block WAL service already started");
             return this;
         }
-        StopWatch stopWatch = StopWatch.createStarted();
+        if (started.compareAndSet(state, true)) {
+            boolean success = false;
+            try {
+                StopWatch stopWatch = StopWatch.createStarted();
 
-        walChannel.open(channel -> Optional.ofNullable(tryReadWALHeader(walChannel))
-            .map(BlockWALHeader::getCapacity)
-            .orElse(null));
+                walChannel.open(channel -> Optional.ofNullable(tryReadWALHeader(walChannel))
+                    .map(BlockWALHeader::getCapacity)
+                    .orElse(null));
 
-        BlockWALHeader header = tryReadWALHeader(walChannel);
-        if (null == header) {
-            assert !recoveryMode;
-            header = newWALHeader();
-            firstStart = true;
-            LOGGER.info("no available WALHeader, create a new one: {}", header);
-        } else {
-            LOGGER.info("read WALHeader from WAL: {}", header);
+                BlockWALHeader header = tryReadWALHeader(walChannel);
+                if (null == header) {
+                    assert !recoveryMode;
+                    header = newWALHeader();
+                    firstStart = true;
+                    LOGGER.info("no available WALHeader, create a new one: {}", header);
+                } else {
+                    LOGGER.info("read WALHeader from WAL: {}", header);
+                }
+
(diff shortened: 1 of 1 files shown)


CLFutureX:
@superhx  @Chillax-0v0  hey, PTAL

s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
```diff
@@ -273,10 +278,34 @@ private void parseRecordBody(long recoverStartOffset, RecordHeader readRecordHea
...
+                        if (!success) {
+                            WAL_STATE.compareAndSet(this, WAL_STATE_STARTED, WAL_STATE_INIT);
+                            LOGGER.warn("block WAL service started fail");
+                        }
+                    }
+                }
+                break;
+            case WAL_STATE_STARTED:
+                LOGGER.warn("block WAL service already started");
+                break;
```
Chillax-0v0, line 298: Currently, the "started" state is primarily used to check whether the metadata required for WAL operation (e.g., WAL Header) is ready (see callers of `checkStarted`), in order to avoid data corruption caused by premature writes. Therefore, the state can only be set to "started" after the `doStart` method executes successfully.
Adding an intermediate "starting" state might resolve the issue.

s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
```diff
@@ -115,7 +116,11 @@ public class BlockWALService implements WriteAheadLog {
     public static final int WAL_HEADER_CAPACITY = WALUtil.BLOCK_SIZE;
     public static final int WAL_HEADER_TOTAL_CAPACITY = WAL_HEADER_CAPACITY * WAL_HEADER_COUNT;
     private static final Logger LOGGER = LoggerFactory.getLogger(BlockWALService.class);
-    private final AtomicBoolean started = new AtomicBoolean(false);
+    private static final int WAL_STATE_INIT = 1;
+    private static final int WAL_STATE_STARTED = 2;
+    private static final int WAL_STATE_SHUTING_DOWN = 3;
+    private static final int WAL_STATE_SHUTDOWN = 4;
+    private static final AtomicIntegerFieldUpdater<BlockWALService> WAL_STATE = AtomicIntegerFieldUpdater.newUpdater(BlockWALService.class, "state");
```
Chillax-0v0, line 123: I think there is no need to use `AtomicIntegerFieldUpdater` here; an `AtomicReference<State>` should suffice. Here are a few reasons:

- In our scenario, there won't be many `WAL` instances simultaneously (in fact, in most cases, there is only one). So a global `static` updater is unnecessary.
- Compared to `AtomicIntegerFieldUpdater`, `AtomicReference` can provide better code readability and maintainability.
- I conducted a simple benchmark, and there is no significant performance difference between them (in fact, for the most frequently called `get` method in our use case, `AtomicReference` performs slightly better).

s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
```diff
@@ -115,7 +116,11 @@ public class BlockWALService implements WriteAheadLog {
     public static final int WAL_HEADER_CAPACITY = WALUtil.BLOCK_SIZE;
     public static final int WAL_HEADER_TOTAL_CAPACITY = WAL_HEADER_CAPACITY * WAL_HEADER_COUNT;
     private static final Logger LOGGER = LoggerFactory.getLogger(BlockWALService.class);
-    private final AtomicBoolean started = new AtomicBoolean(false);
+    private static final int WAL_STATE_INIT = 1;
+    private static final int WAL_STATE_STARTED = 2;
+    private static final int WAL_STATE_SHUTING_DOWN = 3;
+    private static final int WAL_STATE_SHUTDOWN = 4;
```
Chillax-0v0, line 122: If we use `AtomicReference` to maintain the state of `WAL` (as I mentioned below), then these `int` constants can also be replaced with a separate enum class.

Chillax-0v0 (commented):
Hi @CLFutureX,

I apologize for the delay in reviewing this PR. Thank you for your patience and for the effort you've put into this work.

I've gone through the changes and have a few suggestions that I believe will improve the implementation. Could you please consider these modifications?

Thank you again for your contribution!

Chillax-0v0:
@CLFutureX Hello, I have submitted a PR to your branch https://github.com/CLFutureX/automq/pull/1, which converts `WalState` into an enum class. Please take a look.

CLFutureX:
> @CLFutureX Hello, I have submitted a PR to your branch [CLFutureX#1](https://github.com/CLFutureX/automq/pull/1), which converts `WalState` into an enum class. Please take a look.

LGTM

PR #1638, opened 2024-07-24
fix(api): fix node metadata

superhx:


core/src/main/scala/kafka/server/ControllerServer.scala
```diff
@@ -295,11 +295,11 @@ class ControllerServer(
           setDelegationTokenExpiryCheckIntervalMs(config.delegationTokenExpiryCheckIntervalMs).
           setEligibleLeaderReplicasEnabled(config.elrEnabled).
           setStreamClient(streamClient).
+          setExtension(c => quorumControllerExtension(c)).
```
Chillax-0v0, line 298: Missing "AutoMQ for Kafka inject" ?

Chillax-0v0 (approved):

PR #1650, opened 2024-07-25
feat(s3stream): adopt objectStorage throttleStrategy for upload WAL task

lifepuzzlefun:
# what is changed.
1. removed UploadWalTask ratelimiter let object storage do the throttle logic.

# why
1. in our poc environment we remove the limit logic at first. the s3.network.baseline.bandwidth is 1GB/s
2. the upload task cost time drop from 3000ms to 500ms.
3. the failover duration drop from 5s to 1s per partition


lifepuzzlefun:
image

superhx:
Purpose of rate limiting: Smoothing out the uploaded traffic to avoid CPU spikes caused by concentrated uploads, resulting in fluctuating transmission delays.

The current reasons affecting the speed of partition migration: Daily upload tasks have been throttled, blocking the completion of upload tasks triggered by closing Stream.

Solution:
- Add the `#burst` method to DeltaWALUploadTask to ignore the current rate limit of the uploading task and immediately trigger the upload.
- If receiving a DeltaWALUploadTask with force = true, call the `#burst` method of all inflight DeltaWALUploadTasks to accelerate the upload.


https://github.com/AutoMQ/automq/blob/12fb82d2220aa93f1f3add58af7e28499675c3ae/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java#L648-L662

lifepuzzlefun:
> Purpose of rate limiting: Smoothing out the uploaded traffic to avoid CPU spikes caused by concentrated uploads, resulting in fluctuating transmission delays.
> 
> The current reasons affecting the speed of partition migration: Daily upload tasks have been throttled, blocking the completion of upload tasks triggered by closing Stream.
> 
> Solution:
> 
> * Add the `#burst` method to DeltaWALUploadTask to ignore the current rate limit of the uploading task and immediately trigger the upload.
> * If receiving a DeltaWALUploadTask with force = true, call the `#burst` method of all inflight DeltaWALUploadTasks to accelerate the upload.
> 
> https://github.com/AutoMQ/automq/blob/12fb82d2220aa93f1f3add58af7e28499675c3ae/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java#L648-L662

got it, I'll change the code and do the new test

PR #1662, opened 2024-07-29
feat(s3stream): support burst upload for upload WAL task to reduce failover time.

lifepuzzlefun:
# Why this patch
we want to optimize the partition failover time. and we found the main part in partition clean shutdown failover is stream close.

and the stream close is slow because of the Upload RateLimit when upload.
and the FutureTicker also will cause the force upload task finish in long time. which at most delay 500ms and the 500ms delay may cause the upload data amount much larger.

# What is changed
1. support p99, p95 p50 in DeltaHistogram. use HdrHistogram which calculate p99 in low overhead.
2. support DeltaWALUploadTask limiter burst mode which cal make all the acquire fast pass.
3. support FutureTicker to wait small time when force upload.

------

image

kube800 the yellow line is enabled burst and wait at most 100ms when force upload.
kube801 the orange line is enabled burst and wait at most 500ms when force upload.
kube802 the blue line is no burst.

with this patch the s3stream close time p99 can drop from 3s to 500ms in our environment

# test case
3 node each 200MB/s in traffic 30 partition each node. 100 partition total.
submit reassign task each 2min which will change partition leader 100 -> 101, 101 -> 102, 102 -> 100.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
@@ -58,8 +58,6 @@ public class DeltaWALUploadTask {
     private volatile CommitStreamSetObjectRequest commitStreamSetObjectRequest;
     private volatile boolean burst = false;
 
-    public S3Storage.DeltaWALUploadTaskContext ctx;
-
     public DeltaWALUploadTask(Config config, Map<Long, List<StreamRecordBatch>> streamRecordsMap,
                               ObjectManager objectManager, ObjectStorage objectStorage,
                               ExecutorService executor, boolean forceSplit, double rate) {
@@ -153,8 +151,7 @@ public class DeltaWALUploadTask {
                     }
                 }));
             } else {
-                streamSetWriteCfList.add(acquireLimiter(streamSize)
-                    .thenAccept(nil -> streamSetObject.write(streamId, streamRecords)));
+                streamSetWriteCfList.add(acquireLimiter(streamSize).thenAccept(nil -> streamSetObject.write(streamId, streamRecords)));
                 long startOffset = streamRecords.get(0).getBaseOffset();
                 long endOffset = streamRecords.get(streamRecords.size() - 1).getLastOffset();
                 request.addStreamRange(new ObjectStreamRange(streamId, -1L, startOffset, endOffset, streamSize));
--- a/s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
@@ -30,6 +30,9 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.LongAdder;
+import java.util.function.LongSupplier;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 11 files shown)


s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
```diff
@@ -55,10 +58,18 @@ public class DeltaWALUploadTask {
...
+
+    public S3Storage.DeltaWALUploadTaskContext ctx;
+    public LongSupplier burstTaskId;
 
     public DeltaWALUploadTask(Config config, Map<Long, List<StreamRecordBatch>> streamRecordsMap,
-        ObjectManager objectManager, ObjectStorage objectStorage,
-        ExecutorService executor, boolean forceSplit, double rate) {
+                              ObjectManager objectManager, ObjectStorage objectStorage,
+                              ExecutorService executor, boolean forceSplit, double rate,
+                              long id, LongSupplier burstTaskId) {
```
superhx, line 72: Why we need the id and burstTaskId here? DeltaWALUploadTask already has the #burst method

s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
```diff
@@ -55,10 +58,18 @@ public class DeltaWALUploadTask {
     private long uploadTimestamp;
     private long commitTimestamp;
     private volatile CommitStreamSetObjectRequest commitStreamSetObjectRequest;
+    private volatile boolean burst = false;
+    private final long id;
+    private final LongAdder rateLimitTime = new LongAdder();
+    private final LongAdder burstBytes = new LongAdder();
```
superhx, line 64: `rateLimitTime` and `burstBytes` have not been read, is this test code?

s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
```diff
@@ -590,7 +602,7 @@ public CompletableFuture<Void> forceUpload(long streamId) {
         final long startTime = System.nanoTime();
         CompletableFuture<Void> cf = new CompletableFuture<>();
         // Wait for a while to group force upload tasks.
-        forceUploadTicker.tick().whenComplete((nil, ex) -> {
+        forceUploadTicker.tick(enableBurstForceUpload && streamId != LogCache.MATCH_ALL_STREAMS).whenComplete((nil, ex) -> {
```
superhx, line 605: It bypasses the effect of batch uploading, which is equivalent to triggering an upload each time a partition closes. In a scenario with 5000 partitions on a single machine, the effect of batching may be better.

If you think 500ms is too long, you can change the default value to 100ms

s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
lifepuzzlefun: removed.

s3stream/src/main/java/com/automq/stream/s3/DeltaWALUploadTask.java
lifepuzzlefun: removed.

s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
lifepuzzlefun: yeah. when force upload the batch is still applied. but much lower from 500ms to 100ms.
we found that the more batch wait the more data may need to be upload. which also incr the close time.

AUTOMQ_BURST_UPLOAD_DELAY_TICK_TIME config change default to 100ms

superhx (approved):

PR #1699, opened 2024-08-01
feat(s3stream): isolate the wal object from previous epoch

ShadowySpirits:


s3stream/src/main/java/com/automq/stream/s3/wal/impl/object/RecordAccumulator.java
```diff
@@ -297,6 +315,40 @@ public long append(long recordSize, Function<Long, ByteBuf> recordSupplier,
...
+
+        if (objectMap.isEmpty() && previousObjectMap.isEmpty()) {
+            return CompletableFuture.completedFuture(null);
+        }
+
+        long startTime = time.nanoseconds();
+        AtomicLong deletedObjectSize = new AtomicLong();
+        List<ObjectStorage.ObjectPath> deleteObjectList = new ArrayList<>();
+
+        previousObjectMap.forEach((k, v) -> {
```
superhx, line 329: remove in foreach is undefined

ShadowySpirits: In the implementation of this map, removing an element using an iterator is no different from using a foreach loop.

image

superhx (approved):

PR #1717, opened 2024-08-03
feat(s3stream): add storage status check

CLFutureX:
add storage status check

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -56,6 +56,7 @@ import java.util.Optional;
 import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
```diff
@@ -417,7 +419,7 @@ public CompletableFuture<Void> append(AppendContext context, StreamRecordBatch s
      * @return backoff status.
      */
     public boolean append0(AppendContext context, WalWriteRequest request, boolean fromBackoff) {
-        // TODO: storage status check, fast fail the request when storage closed.
+        checkStatus();
```
superhx, line 422: It should be `#completeException(new IOException())`

CLFutureX: yes, that would be better

CLFutureX: > It should be `#completeException(new IOException())`
  yes, that would be better

superhx: It should invoke WalWriteRequest.cf#completeException

Issue #1725, opened 2024-08-06
[Enhancement] optimization Handler chain 


CLFutureX:
The current error handling chain in s3Storage is not distinctly designed to differentiate between the chain and the handlers on the chain, with the distinction being made only in naming, which leads to poor code readability and maintainability.

Optimization: Create an interface for the chain so that the caller can clearly understand that the call is to the chain, not to an individual handler. For example, in the current S3Storage, what needs to be called is the chain, but the parameter passed is the handler.
<img width="748" alt="1722922618367" src="https://github.com/user-attachments/assets/2f8c0b26-a163-4bac-8481-8aeedc55478a">

Bhuvannnn:
Is this still wanted? #1726 added a chain interface and retyped `S3Storage` to take it, then went stale.

One thing from that PR I'm unsure about: @superhx asked why `S3Storage` needs to know it's a `StorageHandlerChain` rather than a single handler. Should the constructor take the chain type, or keep `StorageFailureHandler` and only distinguish the chain at the `DefaultS3Client` wiring site?

Don't want to `/assign` until that's clear.

Bhuvannnn:
cc @superhx @Gezi-lzq — looping you in since this went stale and the #1726 review never landed.

PR #1726, opened 2024-08-06
feat(s3stream): optimization handler chain

CLFutureX:
[issue](https://github.com/AutoMQ/automq/issues/1725)

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -19,7 +19,7 @@ import com.automq.stream.s3.cache.S3BlockCache;
 import com.automq.stream.s3.context.AppendContext;
 import com.automq.stream.s3.context.FetchContext;
 import com.automq.stream.s3.failover.Failover;
-import com.automq.stream.s3.failover.StorageFailureHandler;
+import com.automq.stream.s3.failover.StorageHandlerChain;
 import com.automq.stream.s3.metadata.StreamMetadata;
 import com.automq.stream.s3.metrics.S3StreamMetricsManager;
 import com.automq.stream.s3.metrics.TimerUtil;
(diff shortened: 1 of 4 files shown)


s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
```diff
@@ -128,7 +128,7 @@ public class S3Storage implements Storage {
 
     @SuppressWarnings("this-escape")
     public S3Storage(Config config, WriteAheadLog deltaWAL, StreamManager streamManager, ObjectManager objectManager,
-        S3BlockCache blockCache, ObjectStorage objectStorage, StorageFailureHandler storageFailureHandler) {
```
superhx, line 131: Why S3Storage need to know the storage failure handle is a StorageHandlerChain?

CLFutureX: Based on the current constructor method with the handler parameter, it gives the impression that only a single handler implementation class needs to be executed, whereas in reality, S3Storage requires the execution of multiple handler implementation classes.

CLFutureX: > Why S3Storage need to know the storage failure handle is a StorageHandlerChain?

Based on the current constructor method with the handler parameter, it gives the impression that only a single handler implementation class needs to be executed, whereas in reality, S3Storage requires the execution of multiple handler implementation classes.

CLFutureX: > Why S3Storage need to know the storage failure handle is a StorageHandlerChain?

The root cause lies in the fact that the current design does not distinguish between 'chain' and 'handler' (that is, the chain of responsibility has not been distinguished from the nodes on the chain of responsibility), with differentiation made only in naming.

PR #1732, opened 2024-08-06
feat(core): expose more java modules

lifepuzzlefun:


lifepuzzlefun:
need the java core modules to use arthas. if not enabled the reflection can't use.

superhx (approved):

PR #1779, opened 2024-08-13
feat(s3stream): optimize the execution order of the release method for ByteBuf

CLFutureX:
**Background**:

Due to the fact that the callback order of `CompletableFuture` follows a LIFO (Last In First Out) pattern, it leads to uncertainty and potentially significant delays in the execution timing of the `release` method for the current `StreamRecordBatch`.

We should ensure that such methods are executed with priority whenever possible.
com.automq.stream.s3.S3Storage#append
com.automq.stream.s3.S3Stream#append0


CLFutureX:
@superhx  @Chillax-0v0  hey, please take a look

superhx (approved):

PR #1796, opened 2024-08-14
feat(eventLoop): the visibility of the shutdown flag for the event loop and optimizations for graceful shutdown

CLFutureX:
The visibility of the shutdown flag for the event loop and optimizations for graceful shutdown

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/utils/threads/EventLoop.java
+++ b/s3stream/src/main/java/com/automq/stream/utils/threads/EventLoop.java
@@ -89,6 +89,7 @@ public class EventLoop extends Thread implements Executor {
     }
 
     public synchronized CompletableFuture<Void> shutdownGracefully() {
+        shutdown = true;
         if (!shutdownCf.isDone() && tasks.isEmpty()) {
             tasks.add(WAKEUP_TASK);
         }
--- a/s3stream/src/main/java/com/automq/stream/utils/threads/EventLoop.java
+++ b/s3stream/src/main/java/com/automq/stream/utils/threads/EventLoop.java
@@ -87,6 +87,11 @@ public class EventLoop extends Thread implements Executor {
     public void execute(Runnable task) {
         check();
         tasks.add(task);
+        if (shutdown.get()) {
+            if (tasks.remove(task)) {
+                throw new IllegalStateException("EventLoop is shutdown");
+            }
+        }
     }
 
     public CompletableFuture<Void> shutdownGracefully() {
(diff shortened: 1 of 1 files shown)


s3stream/src/main/java/com/automq/stream/utils/threads/EventLoop.java
```diff
@@ -79,6 +89,9 @@ public void execute(Runnable task) {
 
     public CompletableFuture<Void> shutdownGracefully() {
         shutdown = true;
+        if (!shutdownCf.isDone()) {
```
superhx, line 92: To ensure concurrent safety, the operations of modifying `shutdown` and `tasks` need to be safeguarded by a lock. Similarly, other methods that use these two variables need to be modified.

CLFutureX: > To ensure concurrent safety, the operations of modifying `shutdown` and `tasks` need to be safeguarded by a lock. Similarly, other methods that use these two variables need to be modified.

Yes, perhaps we can modify the shutdown from a constant to an AtomicBoolean, utilizing CAS (Compare-And-Swap) to ensure concurrent safety.

CLFutureX: > To ensure concurrent safety, the operations of modifying `shutdown` and `tasks` need to be safeguarded by a lock. Similarly, other methods that use these two variables need to be modified.

now, the tasks variable has also been modified.

CLFutureX:
@superhx   Perhaps  it should only adjust the visibility of the shutdown variable, without introducing a wakeup mechanism, because the worst-case scenario would only involve a wait of 100ms, which is generally acceptable for a graceful shutdown, What do you think?

s3stream/src/main/java/com/automq/stream/utils/threads/EventLoop.java
```diff
@@ -17,14 +17,21 @@
...
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
+
 import org.slf4j.Logger;
 
 public class EventLoop extends Thread implements Executor {
     private final Logger logger;
     private BlockingQueue<Runnable> tasks;
-    private boolean shutdown = false;
+    private volatile boolean shutdown;
```
superhx, line 26: Did not see when shutdown = true was set

superhx:
> @superhx Perhaps it should only adjust the visibility of the shutdown variable, without introducing a wakeup mechanism, because the worst-case scenario would only involve a wait of 100ms, which is generally acceptable for a graceful shutdown, What do you think?

If it's not completed, I think it's ok to add the shutdown awake.

CLFutureX:
@superhx  hey, please take a look

superhx (approved):

PR #1831, opened 2024-08-16
feat(s3stream): support skip read delta cache when test blockCache logic.

lifepuzzlefun:


s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
```diff
@@ -536,7 +538,13 @@ private CompletableFuture<ReadDataBlock> read0(FetchContext context,
         @SpanAttribute long startOffset,
         @SpanAttribute long endOffset,
         @SpanAttribute int maxBytes) {
-        List<StreamRecordBatch> logCacheRecords = deltaWALCache.get(context, streamId, startOffset, endOffset, maxBytes);
+        List<StreamRecordBatch> logCacheRecords;
+        if (AUTOMQ_SKIP_READ_DELTA_LOG_CACHE) {
```
superhx, line 542: The data may only exist in LogCache, so it cannot be skipped.

superhx: If you want test BlocCache logic, just send data more than LogCache size, and consume from earliest.

The LogCache only cache the WAL data.

PR #1832, opened 2024-08-16
feat(metadata): limit the size of sparse index cache

SCNieh:

PR #1834, opened 2024-08-16
WIP fix(s3stream): fix LocalStreamRangeIndexCache leak when stream destory

lifepuzzlefun:


SCNieh:
@lifepuzzlefun Hi, thank for bring it up, the cleanning of local sparse index cache relies on the result of compaction, when a stream is no longer exists, the stream set objects it belongs to will be compacted into multiple stream objects after 120min, and the corresponding range indexes of the stream will be removed from local cache accordingly

lifepuzzlefun:
@SCNieh hi, I aggree with you. let me explain my test case. create 1w topic partition and delete. the delete will happen within 2 hours. And when force split the stream is gone. so the index won't be cleaned.

SCNieh:
> @SCNieh hi, I aggree with you. let me explain my test case. create 1w topic partition and delete. the delete will happen within 2 hours. And when force split the stream is gone. so the index won't be cleaned.
> 
> In out environment the 3w partition node index size drop from 200MB to 64KB with this patch. coulde you help review ?

1. force split happens per stream set object, and it does not care whether the stream exists or not, so even if the stream is removed, the stream set object it belongs to will still be compacted, as well as its sparse index
2. I just pushed a [PR](https://github.com/AutoMQ/automq/pull/1832) regarding limit the overall size of local sparse index per node, that will help solve the problem you mentioned, the PR is still under review and any comments will be welcomed

lifepuzzlefun:
duplicate with https://github.com/AutoMQ/automq/pull/1856

PR #1880, opened 2024-08-21
perf(s3stream): optimize the critical section in s3 wal

### LINKED ISSUE #1874: [Enhancement] Optimize the performance of WALUtil#generateRecord
The performance of WALUtil#generateRecord, executed in the critical section, will significantly impact latency. We should avoid performing CPU-bound operations.

image

The flame graph indicates that allocating the buffer and calculating the checksum is the most time-consuming task. We could optimize this function by preallocating the buffer and using the more effective checksum algorithm.

ShadowySpirits:
close #1874

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/FixedSizeByteBufPool.java
+++ b/s3stream/src/main/java/com/automq/stream/FixedSizeByteBufPool.java
@@ -66,6 +66,7 @@ public class FixedSizeByteBufPool {
      */
     public void release(ByteBuf buffer) {
         assert buffer.capacity() == bufferSize;
+        assert buffer.refCnt() == 1;
 
         if (poolSize.get() >= maxPoolSize) {
             buffer.release();
--- a/s3stream/src/main/java/com/automq/stream/FixedSizeByteBufPool.java
+++ b/s3stream/src/main/java/com/automq/stream/FixedSizeByteBufPool.java
@@ -67,10 +67,6 @@ public class FixedSizeByteBufPool {
     public void release(ByteBuf buffer) {
         assert buffer.capacity() == bufferSize;
 
-        if (buffer.refCnt() != 1) {
-            throw new IllegalArgumentException("The buffer is not exclusive");
-        }
-
         if (poolSize.get() >= maxPoolSize) {
             buffer.release();
             return;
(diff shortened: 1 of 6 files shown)


s3stream/src/main/java/com/automq/stream/FixedSizeByteBufPool.java
```diff
@@ -67,6 +67,10 @@ private ByteBuf allocate() {
     public void release(ByteBuf buffer) {
         assert buffer.capacity() == bufferSize;
 
+        if (buffer.refCnt() != 1) {
+            throw new IllegalArgumentException("The buffer is not exclusive");
+        }
+
```
Chillax-0v0, line 73: This is a high-frequency method call, so for performance reasons, I think using `assert` here would be sufficient. If a memory leak does occur, Netty's leak detection will also be able to identify the issue.

s3stream/src/main/java/com/automq/stream/s3/wal/impl/object/ObjectWALService.java
```diff
@@ -84,10 +86,33 @@ public WALMetadata metadata() {
...
+                header.release();
+            }
+
+            // Return the header buffer to the buffer pool.
+            if (header.refCnt() == 1) {
+                byteBufPool.release(header);
+            } else {
+                log.error("[Bug] The header buffer is already released.");
+            }
+        });
```
Chillax-0v0, line 103: I think directly checking the `refCnt` of `ByteBuf` is not appropriate. We should ensure that the `ByteBuf` is correctly released through well-structured code, such as handling various exceptions.

s3stream/src/main/java/com/automq/stream/s3/wal/impl/object/ObjectWALService.java
```diff
@@ -36,17 +39,16 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static com.automq.stream.s3.wal.common.RecordHeader.RECORD_HEADER_SIZE;
-import static com.automq.stream.s3.wal.common.RecordHeader.RECORD_HEADER_WITHOUT_CRC_SIZE;
 
 public class ObjectWALService implements WriteAheadLog {
     private static final Logger log = LoggerFactory.getLogger(ObjectWALService.class);
 
+    protected FixedSizeByteBufPool byteBufPool = new FixedSizeByteBufPool(RECORD_HEADER_SIZE, 1024);
```
Chillax-0v0, line 51: When the server is handling a large amount of requests, 1024 headers may not be sufficient. Consider increasing the pool size as the number of CPU cores increases. See #1882 for reference.

Chillax-0v0 (approved):

Chillax-0v0 (approved):

PR #1894, opened 2024-08-22
fix(metadata): fix sparse index leak on compaction split

SCNieh:


s3stream/src/main/java/com/automq/stream/s3/objects/StreamObject.java
```diff
@@ -18,7 +18,24 @@ public class StreamObject {
...
     private long startOffset;
     private long endOffset;
-    private int attributes = ObjectAttributes.UNSET.attributes();
+    private int attributes;
+
+    public StreamObject() {
+        this(0L, 0L, 0L, 0L, 0L);
+    }
+
+    public StreamObject(long objectId, long objectSize, long streamId, long startOffset, long endOffset) {
```
superhx, line 27: The test constructor should only add in test code. The StreamObject should keep the full fields constructor to avoid mistaking init.

s3stream/src/main/java/com/automq/stream/s3/index/LocalStreamRangeIndexCache.java
```diff
@@ -310,10 +326,10 @@ public CompletableFuture<Void> updateIndexFromRequest(CommitStreamSetObjectReque
             }
             rangeIndexMap.put(range.getStreamId(), newRangeIndex);
         }
-        if (request.getCompactedObjectIds().isEmpty()) {
-            return append(rangeIndexMap);
+        for (StreamObject streamObject : request.getStreamObjects()) {
```
superhx, line 329: If it isn't a compact request, the stream object will be record

superhx (approved):

PR #1897, opened 2024-08-22
fix(s3stream/wal): fix memory leak

Chillax-0v0:
The same as #1896


s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockImpl.java
```diff
@@ -93,7 +97,7 @@ public long addRecord(long recordSize, RecordSupplier recordSupplier,
 
         long recordOffset = startOffset + nextOffset;
         recordSuppliers.add(() -> {
-            ByteBuf header = HEADER_POOL.get();
+            ByteBuf header = HEADER_POOL.get().retain();
```
superhx, line 100: When the header is released. Base on ownership principle, when HEADER_POOL#get has already transfer the  ownership of ByteBuf to outside.

superhx (approved):

PR #1932, opened 2024-08-23
feat(metadata): blocking wait for uploading local index cache on stre…

SCNieh:
…am close


s3stream/src/main/java/com/automq/stream/s3/index/LocalStreamRangeIndexCache.java
```diff
@@ -120,7 +115,7 @@ public CompletableFuture<Void> upload() {
 
     private void batchUpload() {
         batchUpload0().whenComplete((v, ex) -> {
-            executorService.schedule(this::batchUpload, 10, TimeUnit.MILLISECONDS);
+            executorService.schedule(this::batchUpload, 1000, TimeUnit.MILLISECONDS);
```
lifepuzzlefun, line 118: I think 10ms is better, worst case wait 1s will increase failover time.

SCNieh: yes, that's a test code mistakenly get pushed

superhx (approved):

PR #1941, opened 2024-08-26
feat(s3stream): support limit write bandwidth for WALBlockDeviceChannel

lifepuzzlefun:
### why need this?
1. the block device provider have limit on iobandwidth and iops. either one exceed the limit the write block channel time goes very high (cause by the block device qos)
2. when single node write request number is high. the write may exceed the limit iobandwidth for a little time but can cause the pending request accumlate and finally the broker will have very long write latency.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/wal/util/WALBlockDeviceChannel.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/wal/util/WALBlockDeviceChannel.java
@@ -16,10 +16,13 @@ import com.automq.stream.s3.wal.exception.WALNotInitializedException;
 import com.automq.stream.thirdparty.moe.cnkirito.kdio.DirectIOLib;
 import com.automq.stream.thirdparty.moe.cnkirito.kdio.DirectIOUtils;
 import com.automq.stream.thirdparty.moe.cnkirito.kdio.DirectRandomAccessFile;
+import io.github.bucket4j.BlockingBucket;
+import io.github.bucket4j.Bucket;
 import io.netty.buffer.ByteBuf;
 import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.time.Duration;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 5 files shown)


Chillax-0v0:
Hello, thank you for your contribution. However, could you please explain the reasons for doing so?

lifepuzzlefun:
image
our disk the upper limit is 210MB/s and 2800 iops. but the incoming request size maybe larger then one block so the iops not exceed but the write bandwidth exceeded.

s3stream/src/main/java/com/automq/stream/s3/wal/util/WALBlockDeviceChannel.java
```diff
@@ -285,6 +299,15 @@ private void unalignedWrite(ByteBuf src, long position) throws IOException {
     private int write(ByteBuffer src, long position) throws IOException {
         assert WALUtil.isAligned(src.remaining());
 
+        int permits = src.remaining();
+        if (writeBandwidthLimit != DISABLE_WRITE_BAND_WIDTH_LIMIT && permits > 0) {
```
superhx, line 303: Default use Long.MAX_VALUE? Reduce branching logic.

lifepuzzlefun:
code update, notice that bucket4j limit max 1token/ns which is 953.67MB/s, and when acquire 0 exception will be throwed.

superhx:
> code update, notice that bucket4j limit max 1token/ns which is 953.67MB/s, and when acquire 0 exception will be throwed.

Should we throw exception when the bandwidth config exceeds the limitation or support higher limitation?

lifepuzzlefun:
> > code update, notice that bucket4j limit max 1token/ns which is 953.67MB/s, and when acquire 0 exception will be throwed.
> 
> Should we throw exception when the bandwidth config exceeds the limitation or support higher limitation?

@superhx code changed to support 3814.7 MB/s. and when consume the bytes will >> 4. when user config larger than this exception will throw.

superhx (approved):

PR #1951, opened 2024-08-27
perf(core): disable DeleteProduce tryComplete trace log

lifepuzzlefun:
1. when broker under high load the pending completed produce request is about 20K.
2. the appendCallback check tryComplete which traceLog consume almost 20% cpu.


lifepuzzlefun:
image

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1190,8 +1190,10 @@ class Partition(val topicPartition: TopicPartition,
       case Some(leaderLog) =>
         // keep the current immutable replica list reference
         val curMaximalIsr = partitionState.maximalIsr
-
-        if (isTraceEnabled) {
```
superhx, line 1194: It already has traceEnabled check, why they will cost CPU

lifepuzzlefun: this isTraceEnabled is a method which will query the log4j or log implementation to check the log level 
image
when high rate check tryComplete the method will burst cpu.

lifepuzzlefun: this patch only disable the two part trace check with call the method to check the log implementation log level.
image

superhx: Could we check isTraceEnabled and cache the result instead of introduce AUTOMQ_ENABLE_TRACE_LOG switch

superhx (approved):

PR #1952, opened 2024-08-27
feat(wal): unOverCapacityException will lead to memory leaks

CLFutureX:
Background：
When it is not an OverCapacityException, it can cause memory leaks.
path: com.automq.stream.s3.wal.impl.block.BlockWALService#append


s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
```diff
@@ -421,11 +421,13 @@ public AppendResult append(TraceContext context, ByteBuf buf, int crc) throws Ov
             AppendResult result = append0(buf, crc);
             result.future().whenComplete((nil, ex) -> TraceUtils.endSpan(scope, ex));
             return result;
-        } catch (OverCapacityException ex) {
+        } catch (Throwable t) {
             buf.release();
-            StorageOperationStats.getInstance().appendWALFullStats.record(TimerUtil.timeElapsedSince(startTime, TimeUnit.NANOSECONDS));
-            TraceUtils.endSpan(scope, ex);
-            throw ex;
+            if (t instanceof OverCapacityException) {
```
superhx, line 426: The record could be removed

CLFutureX: > The record could be removed

I wonder where it will be removed? Is the release method called when it is removed?
I understand that when the BlockWALService#append method throws a non-OverCapacityException exception, at this time buf's refCnt=4,
the external S3Storage#append will call the release method once, at this time buf's refCnt=2, its failure to trigger the reduction of the reference count of the huge buf.

superhx: image

superhx (approved):

PR #1966, opened 2024-08-30
perf(core): optimize FetchSession.update high memory allocate rate

lifepuzzlefun:


lifepuzzlefun:
before 8% memory allocation
image

after 0.83 % memory allocation
image

superhx (approved):

PR #1973, opened 2024-08-30
fix(log): use the same view to calculate trim offsets

Chillax-0v0:
The same as #1972


core/src/main/scala/kafka/log/streamaspect/ElasticLogSegmentManager.java
```diff
@@ -93,35 +135,49 @@ public CompletableFuture<ElasticLogMeta> asyncPersistLogMeta() {
...
+    }
 
-        streamManager.streams().forEach((streamName, stream) -> {
-            var minOffset = streamMinOffsets.get(streamName);
+    /**
+     * Calculate trim offset of each stream.
+     *
+     * @return stream trim offset map, key is stream name, value is trim offset.
+     */
+    private static Map<String, Long> calTrimOffset(Map<String, Stream> streams,
```
superhx, line 160: The calTrimOffset should be called before the callback.

superhx: And we should merge calTrimOffset and calStreamsMinOffset to one method

superhx (approved):

PR #2049, opened 2024-10-09
feat(metadata): change KVImage state to TimelineHashMap

lifepuzzlefun:
when the broker local want to access the KV image to get metaStreamKey for topic.
the result may return null. change to TimelineHashMap the key can read successfully


superhx:
Why change the data structure will work. Could you provide more information?

lifepuzzlefun:
for honest I didn't figure out why the read from KVImage return null. current data structure works because the state will be stored in the timelineHashmap the same as kraft metadata KVControllerManager

I have one trace log use arthas to find that the DeltaMap lowest level have the key I want to get but the result return null.

image
image

PR #2203, opened 2024-12-02
feat(config): add table topic conversion type configuration

Gezi-lzq:


storage/src/main/java/org/apache/kafka/storage/internals/log/LogConfig.java
```diff
@@ -723,6 +727,13 @@ public static void validate(Properties props,
         }
     }
 
+    public static void validateTableTopicSchemaConfigValues(Properties props, String tableTopicSchemaRegistryUrl) {
```
superhx, line 730: inject start end

superhx (approved):

PR #2261, opened 2025-01-08
refactor(controller): consider brokers that has recently `CONTROLLED_SHUTDOWN` as `SHUTTING_DOWN`

Chillax-0v0:


metadata/src/main/java/org/apache/kafka/controller/stream/NodeState.java
```diff
@@ -12,5 +12,30 @@
...
+    SHUTTING_DOWN,
+    /**
+     * The node is shut down and cannot handle requests.
+     */
+    SHUTDOWN,
+    /**
+     * Use @{@link #SHUTTING_DOWN} instead.
+     */
+    @Deprecated
+    CONTROLLED_SHUTDOWN,
```
superhx, line 31: Remove useless state

superhx (approved):

PR #2335, opened 2025-03-07
perf(s3stream): limit write traffic to object storage

Chillax-0v0:


s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
```diff
@@ -596,6 +598,32 @@ private void continuousCheck(List<StreamRecordBatch> records) {
         }
     }
 
+    private void maybeForceUpload() {
+        // Limit the number of inflight force upload tasks to avoid too many S3 objects.
+        if (hasInflightForceUploadTask()) {
```
superhx, line 603: The inflightWALUploadTasks and needForceUpload may be accessed by multiple threads

superhx (approved):

Issue #2303, opened 2025-02-14
[Good first issue] Exception while processing response


superhx:
## Error Log
```
2025-02-14 09:14:38.114 +0000 ERROR [kafka.network.Processor]  Exception while processing response for 10.132.46.173:9092-10.221.143.9:40948-238589
java.lang.IllegalStateException: Attempt to retrieve channel for which there is no connection. Connection id 10.132.46.173:9092-10.221.143.9:40948-238589 existing connections [...]
	at org.apache.kafka.common.network.Selector.openOrClosingChannelOrFail(Selector.java:1005)
	at org.apache.kafka.common.network.Selector.unmute(Selector.java:771)
	at kafka.network.Processor.processNewResponses(SocketServer.scala:1090)
	at kafka.network.Processor.run(SocketServer.scala:1016)
```

## Expect
- Check whether the log action is reasonable.
- If it's reasonable, then degrade it to WARN level.

decadezzz:
could you please assign this task to me?

fjhdream:
## Description
There is a proposal to downgrade the error level in Selector.openOrClosingChannelOrFail method from throwing an IllegalStateException to a WARN level log. After careful analysis, I believe this change would be inappropriate for the following reasons:
## Analysis
1. Based on the error call stack, this issue occurs when processing an EndThrottlingResponse in the processNewResponses() method.
2. This error can only be triggered if a StartThrottlingResponse has been previously processed for the same channel.
3. This suggests an abnormal situation where the KafkaChannel has been prematurely removed from the channels map, which indicates a potential bug or race condition in the code.
## Reasoning
Downgrading this error to a WARN level would hide a legitimate issue that should be addressed. The IllegalStateException correctly indicates that the system has reached an inconsistent state:
A channel was throttled (via StartThrottlingResponse)
Later, when trying to unthrottle it (via EndThrottlingResponse), the channel no longer exists
This sequence should not happen in normal operation.

Seed123-psy:
**Recommendation:**
This IllegalStateException should be downgraded to WARN level because it reflects normal connection lifecycle management in Kafka.
**Key Reasons:**
1.Normal Connection Termination Scenarios
Occurs when:
• Client actively closes the connection
• Network instability causes abrupt disconnects
Broker attempts to process a connection already in CLOSING/CLOSED state
Common in distributed systems with transient connections

2.No Functional Impact
• Does NOT affect:
• Message durability or delivery guarantees
[...]

PR #2354, opened 2025-03-15
fix(network): Downgrade IllegalStateException log level from ERROR to…

wangyantong2000:
## Error Log
```
2025-02-14 09:14:38.114 +0000 ERROR [kafka.network.Processor]  Exception while processing response for 10.132.46.173:9092-10.221.143.9:40948-238589
java.lang.IllegalStateException: Attempt to retrieve channel for which there is no connection. Connection id 10.132.46.173:9092-10.221.143.9:40948-238589 existing connections [...]
	at org.apache.kafka.common.network.Selector.openOrClosingChannelOrFail(Selector.java:1005)
	at org.apache.kafka.common.network.Selector.unmute(Selector.java:771)
	at kafka.network.Processor.processNewResponses(SocketServer.scala:1090)
	at kafka.network.Processor.run(SocketServer.scala:1016)
```
## Log rationality analysis:
When processing responses, it is normal for the client to disconnect and not log as an ERROR level.
## Downgrading plan:
Modify the log level to more accurately control the log level of specific exceptions without affecting other logs that require an ERROR level

### DIFF
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -1066,17 +1066,22 @@ private[kafka] class Processor(
             // AutoMQ for Kafka inject start
             val channelContext = channelContexts.get(channelId)
             val channel = selector.channel(channelId)
-            if (channel != null && channel.isMuted) {
-              val unmute = if (channelContext == null) {
-                true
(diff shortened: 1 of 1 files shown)


superhx:
close https://github.com/AutoMQ/automq/issues/2303

superhx (approved):

PR #2382, opened 2025-03-26
fix(AsyncSemaphore): prevent potential semaphore resource leakage when task throw exception in acquire

Gallopm:
When the semaphore has sufficient permits, task will acquire the `requiredPermits`, and
begin executing. Under normal circumstances, task.get() returns `CompletableFuture` and
then registers `whenComplete` method on it. When the task is truly completed, `release`
will be automatically called to release the acquired permits. However, when task.get()
throws exception, the whenComplete method will not be successfully registered, and the
permits acquired by the task will never be automatically released, resulting in semaphore
resource leakage.

This commit explicitly release the `requiredPermits` when `task.get()` throw exception,
avoiding potential permits leak.


Chillax-0v0:
@superhx Mind reviewing this PR?

superhx (approved):

Issue #2374, opened 2025-03-18
[Good First Issue] Support Mount-Based Local Development Setup for AutoMQ with Docker Compose


Chillax-0v0:
### Background
AutoMQ currently provides a simplified Docker Compose and LocalStack-based setup for local development (see [docker-compose.yaml](https://github.com/AutoMQ/automq/blob/main/docker/docker-compose.yaml)). However, this approach relies on pre-built AutoMQ Docker images, which require significant time to build and are inefficient for iterative development and debugging.

To streamline the workflow, we want to **add an alternative Docker Compose-based setup** that allows developers to:
1. Compile Java code locally.
2. Mount the compiled output and relevant paths into a Docker container (using a fixed/base image).
3. Run AutoMQ directly from the mounted paths for real-time testing and debugging.

### Required Changes
Design and implement a Docker Compose configuration that:
1. Uses a **fixed/base Docker image** (e.g., OpenJDK or a lightweight runtime image).
2. Mounts locally compiled Java artifacts (e.g., JAR files, configuration, or classpaths) into the container.
3. Ensures AutoMQ can start successfully using the mounted code without rebuilding the Docker image.

JanYork:
Is it necessary to package the environment into docker, and the compiled product of AutoMQ exists locally. When the product changes, the container will read the compiled product of AutoMQ on the host and run it?

Chillax-0v0:
> Is it necessary to package the environment into docker, and the compiled product of AutoMQ exists locally. When the product changes, the container will read the compiled product of AutoMQ on the host and run it?

The container doesn't need auto-detection. Developers will manually restart it after rebuilding locally. This keeps the setup simple while preserving Docker's environment benefits.

shawngao-org:
> > Is it necessary to package the environment into docker, and the compiled product of AutoMQ exists locally. When the product changes, the container will read the compiled product of AutoMQ on the host and run it?
> 
> The container doesn't need auto-detection. Developers will manually restart it after rebuilding locally. This keeps the setup simple while preserving Docker's environment benefits.

Hello, I have a question:
For this requirement, can I use a Shell script to dynamically build a separate `docker-compose.yaml` file, and then use this orchestration file to create and run the container? (The image currently uses `automqinc/automq:1.3.0`)

1. I feel that the location of the built products is a bit scattered.
2. It seems that it is not possible to just use the `docker-compose.yaml` file? (If you have any objections, please raise them)

What do you think?
[...]

PR #2388, opened 2025-03-27
feat: More convenient local code testing

lingting:
impl #2374 

fast run local code on docker. 
steps:
1. `./gradlew releaseTarGz -x test -x check` 
2. docker-compose -f docker/local/docker-compose.yml up -d

### DIFF
--- a/docker/local/docker-compose.yml
+++ b/docker/local/docker-compose.yml
@@ -35,6 +35,24 @@ services:
     networks:
       - automq_net
 
+# Only comment out, do not remove. For easier future testing
+#  test:
+#    container_name: test
+#    hostname: test
+#    #    image: amazoncorretto:17.0.14
+#    #    image: gradle:jdk17
+#    image: eclipse-temurin:17-jdk-noble
+#    working_dir: /opt/automq
+#    volumes:
+#      - ../../core/build/distributions:/opt/volume_libs:ro
+#      - ../scripts:/opt/volume_scripts:ro
+#    command:
+#      - bash
+#      - -c
+#      - |
+#        /opt/volume_scripts/pre_start.sh && \
+#        tail -f /dev/null
+
   controller:
     container_name: "${CONTROLLER_DOCKER_NAME-controller}"
     hostname: "${CONTROLLER_DOCKER_NAME-controller}"
@@ -45,7 +63,7 @@ services:
       - KAFKA_S3_SECRET_KEY=test
       - KAFKA_HEAP_OPTS=-Xms1g -Xmx1g -XX:MetaspaceSize=96m
     volumes:
-      - ../../core/build/local:/opt/volume_libs:ro
+      - ../../core/build/distributions:/opt/volume_libs:ro
       - ../scripts:/opt/volume_scripts:ro
     command:
       - bash
@@ -78,7 +96,7 @@ services:
       - KAFKA_CFG_ADVERTISED_LISTENERS=PLAINTEXT://broker1:9092,EXTERNAL://localhost:9094
       - KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP=CONTROLLER:PLAINTEXT,EXTERNAL:PLAINTEXT,PLAINTEXT:PLAINTEXT
     volumes:
-      - ../../core/build/local:/opt/volume_libs:ro
+      - ../../core/build/distributions:/opt/volume_libs:ro
       - ../scripts:/opt/volume_scripts:ro
     command:
       - bash
@@ -112,7 +130,7 @@ services:
       - KAFKA_CFG_ADVERTISED_LISTENERS=PLAINTEXT://broker2:9092,EXTERNAL://localhost:9095
       - KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP=CONTROLLER:PLAINTEXT,EXTERNAL:PLAINTEXT,PLAINTEXT:PLAINTEXT
     volumes:
-      - ../../core/build/local:/opt/volume_libs:ro
+      - ../../core/build/distributions:/opt/volume_libs:ro
       - ../scripts:/opt/volume_scripts:ro
     command:
       - bash
--- a/docker/README.md
+++ b/docker/README.md
@@ -122,7 +122,7 @@ Use local code run in docker
 1. generate tgz
 ```shell
 # For example only, can be modified based on your compilation requirements
-./gradlew buildLocal -x test -x check
+./gradlew releaseTarGz -x test -x check
 ```
 2. run
 ```shell
@@ -132,7 +132,7 @@ docker-compose -f docker/local/docker-compose.yml up -d
 - After modifying your code, simply regenerate the tgz and restart the specified service.
 ```shell
 # For example only, can be modified based on your compilation requirements
-./gradlew buildLocal -x test -x check
+./gradlew releaseTarGz -x test -x check
 # eg: restart broker
 docker-compose -f docker/local/docker-compose.yml up broker1 broker2 -d --force-recreate 
 ```
--- a/docker/scripts/pre_start.sh
+++ b/docker/scripts/pre_start.sh
@@ -1,15 +1,14 @@
 #!/usr/bin/env bash
 
-echo "[PreStart] shell check"
-(hostname -I | awk '{print $1}') || exit 1
 echo "[PreStart] mkdir"
 rm -rf /opt/automq
 mkdir -p /opt/kafka || exit 1
 ln -s /opt/kafka /opt/automq || exit 1
-echo "[PreStart] copy"
-cp -r /opt/volume_libs /opt/kafka/kafka || exit 1
+echo "[PreStart] file"
+for f in /opt/volume_libs/*.tgz; do
+  tar -xzf "$f" -C /opt/kafka --one-top-level=kafka --strip-components=1 --overwrite
+done
 cp -r /opt/volume_scripts /opt/kafka/scripts || exit 1
-echo "[PreStart] chmod"
 find /opt/kafka -type f -name "*.sh" -exec chmod a+x {} \;
 echo "[PreStart] env"
 echo "export DEBIAN_FRONTEND=noninteractive" >> ~/.bashrc


Chillax-0v0 (approved):
LGTM

Chillax-0v0:
@superhx, @SCNieh, @Gezi-lzq, Could you please review this PR when you have a moment? Do you have any additional feedback or suggestions?

docker/local/docker-compose.yml
```diff
@@ -0,0 +1,160 @@
...
+    volumes:
+      - ../../core/build/distributions:/opt/volume_libs:ro
+      - ../scripts:/opt/volume_scripts:ro
+    command:
+      - bash
+      - -c
+      - |
+        cp -r /opt/volume_scripts/ /opt/automq/scripts && chmod +x /opt/automq/scripts/*.sh
+        /opt/automq/scripts/pre_start.sh -s /opt/volume_libs -t /opt/kafka && \
+        tail -f /dev/null
```
Chillax-0v0, line 51: Test code mistakenly committed.

docker/local/docker-compose.yml
```diff
@@ -0,0 +1,160 @@
...
+      - automq_net
+    depends_on:
+      - localstack
+      - aws-cli
+
+  broker1:
+    container_name: "${BROKER1_DOCKER_NAME-broker1}"
+    hostname: "${BROKER1_DOCKER_NAME-broker1}"
+    stop_grace_period: 2m
+    image: automqinc/automq-basic:2025.03.27
```
Gezi-lzq, line 82: Consider using a more descriptive tag than the time-based version. Additionally, is it possible to leverage existing pre-built images as the base image here?

docker/local/docker-compose.yml
lingting: > Test code mistakenly committed.

removed

docker/local/docker-compose.yml
lingting: > Consider using a more descriptive tag than the time-based version. Additionally

Initially, I considered using version numbers aligned with the project, but later realized this base image wouldn’t change frequently. So I opted for timestamps as version tags. Are there better suggestions? Or should I just use ‘0.0.1’ as the version number?

> is it possible to leverage existing pre-built images as the base image here?

I opted for manual image building instead of using pre-existing images because the docker/Dockerfile follows this construction approach. To mitigate potential runtime issues, I chose to perform the build manually.

Gezi-lzq: I'm not entirely clear on the implementation details here. I just want to confirm if there's a need for a dedicated automq-basic image in the future. If we use a pre-released image and replace/mount the lib directory with local files at runtime, would this be feasible? I’d like to hear your thoughts.

Chillax-0v0: @Gezi-lzq mentioned that adding an additional base image would increase future maintenance overhead. 

@lingting, My thought is: Could we consider directly using the OpenJDK Docker image as the base image, and perform operations like setting environment variables and creating directories during the startup phase instead?

lingting: @Gezi-lzq  @Chillax-0v0 

The reason I maintain a separate automq-basic image is that I noticed line in the `docker/Dockerfile` .

```
RUN apt-get update && apt-get install -y ca-certificates dnsutils iputils-ping curl wget netcat-openbsd sysstat net-tools htop procps zlib1g vim less openjdk-17-jdk libjemalloc-dev ipset iproute2 && \
    apt-get clean && rm -rf /var/lib/apt/lists /var/cache/apt/archives
```

If the pre-built OpenJDK Docker image doesn't include these required dependencies, installing them during startup would result in excessively long launch times each time.

However, the additional maintenance burden introduced by automq-basic is also a concern. I will now try exploring OpenJDK Docker images to see if there's a suitable version that meets our requirements, allowing us to use it directly without needing to install dependencies during startup.

Chillax-0v0: As far as I know, among the listed dependencies, only `openjdk-17-jdk` and `libjemalloc-dev` are actually required for AutoMQ (in fact, `libjemalloc-dev` is only installed to achieve better performance and isn't necessary for local development purposes).
The other dependencies like `sysstat` and `net-tools` are purely installed for operational debugging convenience, which I believe are unnecessary to install.

lingting: @Gezi-lzq  @Chillax-0v0 
OpenJdk Image is  `Pre-release / non-production builds of OpenJDK` 
`amazoncorretto`  There is no Debian-based image, perhaps we really need to maintain an autumq-basic. However, we can make the docker/Dockerfile build based on automq-basic, so that we no longer need to maintain docker/Dockerfile. Only docker/basic/Dockerfile needs to be maintained.

The new docker/Dockerfile might look like this:
```
FROM docker.io/automqinc:automq-basic:2025.03.27

MAINTAINER AutoMQ for Apache Kafka dev@automq.com

COPY "./automq-*.tgz" /opt/kafka/kafka.tgz
RUN tar -xzvf "/opt/kafka/kafka.tgz" -C /opt/kafka/ --one-top-level=kafka --strip-components 1 && rm "/opt/kafka/kafka.tgz"
COPY scripts /opt/kafka/scripts
RUN chmod +x /opt/kafka/scripts/*.sh
```

Our existing shell scripts are probably written for Debian, right? I tried using multiple images like amazoncorretto:17.0.14-alpine3.21 and amazoncorretto:17.0.14-al2, but none of them could start properly.

Chillax-0v0: I’m not entirely sure if our current shell scripts are only compatible with Debian OS. Out of curiosity, could you share which specific shell script failed and what errors occurred when running it on other OSes?

Additionally, I noticed @shawngao-org used the `gradle:jdk17` image in another PR #2390. I haven’t actually tested it myself, but if possible, could you give it a try?

Finally, regarding your proposal to unify the base image—this is also a valid approach (in fact, @Gezi-lzq had the same idea). If the previous methods don’t work out, this could be a good way to go.

lingting: @Chillax-0v0 
> I’m not entirely sure if our current shell scripts are only compatible with Debian OS. Out of curiosity, could you share which specific shell script failed and what errors occurred when running it on other OSes?

Apologies, I mistakenly equated bash with Debian. It's just that all our current scripts are based on bash.  

@Gezi-lzq 
Now that we're using `eclipse-temurin:17-jdk-noble` as the base image, maintaining `automq-basic` is no longer necessary.

Chillax-0v0:
Hello, @lingting, once the PR is ready for review, please at me.

lingting:
> Hello, @lingting, once the PR is ready for review, please tell me.

@Chillax-0v0  Hi, It has been revised and is ready for further review.

build.gradle
Chillax-0v0, line 1: To minimize the workload when merging Apache Kafka code, keep modifications to the original Kafka code to a minimum, and especially avoid formatting the original Kafka code (even though there are indeed formatting issues).

Gezi-lzq (approved):

build.gradle
lingting: > To minimize the workload when merging Apache Kafka code, keep modifications to the original Kafka code to a minimum, and especially avoid formatting the original Kafka code (even though there are indeed formatting issues).

Let me try a different approach to achieve the goal without modifying build.gradle.

lingting:
@Chillax-0v0 Hi, no longer modifying build.gradle. Please review it again.

shawngao-org:
I went through the submitted code. It seems that you are using the built tarball to "create/simulate" the layer of AutoMQ's original image. As @Chillax-0v0  said:

> A practical workaround is to mount the entire root build directory into the container. This avoids cherry-picking specific artifact paths while maintaining compatibility with Kafka’s native build system.

This may be a bit complicated just for running and debugging purposes. (Or does this include some other functionality? If yes, you can ignore this message.)

I'm not sure if this would violate the KISS principle. (I may need to ask some advice here @Chillax-0v0 )

I'm not sure my idea is correct either. Of course, this is my personal opinion and any objections may be raised.

lingting:
> I went through the submitted code. It seems that you are using the built tarball to "create/simulate" the layer of AutoMQ's original image. As @Chillax-0v0 said:
> 
> > A practical workaround is to mount the entire root build directory into the container. This avoids cherry-picking specific artifact paths while maintaining compatibility with Kafka’s native build system.
> 
> This may be a bit complicated just for running and debugging purposes. (Or does this include some other functionality? If yes, you can ignore this message.)
> 
> I'm not sure if this would violate the KISS principle. (I may need to ask some advice here @Chillax-0v0 )
> 
> I'm not sure my idea is correct either. Of course, this is my personal opinion and any objections may be raised.

If the entire project path is mapped into the container, the following issues will arise:  

1. In the `pre_start.sh` script, I need to integrate existing resources and then distribute them according to the directory hierarchy required by `Kafka` for startup. Otherwise, Kafka cannot start properly. This is a significant task and still requires executing the `releaseTarGz` command before startup to avoid issues with uncompiled resources.  

2. If Kafka adjusts the locations of some resources, I will obviously need to readapt the `pre_start.sh` script accordingly.  

On the other hand, using the output of `releaseTarGz` directly avoids the above problems and ensures better compatibility with potential future changes in Kafka.

Chillax-0v0:
@shawngao-org @lingting Thank you both for the thoughtful discussion! You've raised valid points. To clarify the context: 

Originally, when this issue was raised, the idea was to directly mount the entire AutoMQ directory into the container (after running `./gradlew jar -x test` to compile the source) and then start the Kafka process. As @lingting rightly pointed out, it introduces challenges like managing directory hierarchy compatibility in `pre_start.sh` and potential fragility against Kafka's internal path changes.

While eliminating `releaseTarGz` (as @shawngao-org suggested) would indeed be ideal for faster iteration, the current PR's tarball-based approach provides a more deterministic and stable interface for now. This aligns with Kafka's native artifact expectations and avoids runtime path reconciliation risks. 

Let’s merge this PR to unblock immediate needs, but I fully agree there’s room to optimize the workflow further. Future efforts could explore hybrid approaches (e.g., selective artifact mounting without full tarball rebuilds). Contributions to this end are very welcome—feel free to prototype and iterate!

Appreciate the healthy debate here—this is exactly how we balance pragmatism and long-term maintainability.

Chillax-0v0 (approved):

Issue #2393, opened 2025-03-31
[BUG] WAL upload method gets negative stream size which results in guava RateLimiter throwing exception and kafka can't startup


keashem:
### What went wrong?
when broker recovers from crash, it loads WAL's data; but in some case, the datas in wal not upload to s3 storage is too large, which results in  WAL upload method getting negative stream size , and guava RateLimiter throws exception and kafka can't startup ,the log info as follow:

[2025-03-30 10:43:34,149] INFO try recover from crash, recover records bytes size 4287034156 (com.automq.stream.s3.S3Storage)
[2025-03-30 10:43:34,194] ERROR upload run with unexpected exception (com.automq.stream.s3.DeltaWALUploadTask)
java.lang.IllegalArgumentException: Requested permits (-326332735) must be positive
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:191)
	at com.google.common.util.concurrent.RateLimiter.checkPermits(RateLimiter.java:496)
	at com.google.common.util.concurrent.RateLimiter.tryAcquire(RateLimiter.java:413)
	at com.google.common.util.concurrent.RateLimiter.tryAcquire(RateLimiter.java:368)
	at com.automq.stream.utils.AsyncRateLimiter.acquire(AsyncRateLimiter.java:43)
	at com.automq.stream.s3.DeltaWALUploadTask.acquireLimiter(DeltaWALUploadTask.java:112)
	at com.automq.stream.s3.DeltaWALUploadTask.upload0(DeltaWALUploadTask.java:147)
	at com.automq.stream.s3.DeltaWALUploadTask.lambda$upload$1(DeltaWALUploadTask.java:116)
	at com.automq.stream.utils.FutureUtil.exec(FutureUtil.java:89)
	at com.automq.stream.s3.DeltaWALUploadTask.lambda$upload$2(DeltaWALUploadTask.java:116)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:718)
	at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:482)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
[...]

PR #2394, opened 2025-03-31
fix(s3stream): fix upload stream size overflow

keashem:
WAL upload method gets negative stream size , which results in guava RateLimiter throwing exception and kafka can't startup
[issue link
](https://github.com/AutoMQ/automq/issues/2393)


superhx:
closed https://github.com/AutoMQ/automq/issues/2393

PR #2408, opened 2025-04-07
docs: Update IDE launch configuration in contribution guide

shawngao-org:
In the JDK virtual machine parameters, `-Xms` is greater than `-Xmx`, which will generate an error of `Initial heap size set to a larger value than the maximum heap size`.


Chillax-0v0 (approved):
LGTM

PR #2420, opened 2025-04-16
fix: using SecureRandom in perf tools

yannaingtun:
**Description**
This PR fixes a potential security vulnerability in the randomString method in PerfConfig.java by replacing ThreadLocalRandom with SecureRandom for generating random strings.

**Issue**
The current implementation uses ThreadLocalRandom which is not cryptographically secure and could lead to predictable tokens when used for security purposes.

**Fix**
Replace ThreadLocalRandom.current() with new SecureRandom() to ensure cryptographic strength.

**References**
Similar to vulnerability in CVE-2017-15911
Fixed in Openfire: https://github.com/igniterealtime/Openfire/commit/7ff1f73077cca1b3ec25d5897f31f15340c8fa4e

### DIFF
--- a/tools/src/main/java/org/apache/kafka/tools/automq/perf/PerfConfig.java
+++ b/tools/src/main/java/org/apache/kafka/tools/automq/perf/PerfConfig.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
+import java.security.SecureRandom;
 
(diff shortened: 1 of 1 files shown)


SCNieh:
Hi, thanks for your commit, however the purpose of using Random here is simply to generate non-repeatable topic names for benchmark test, so I don't think is neccessary to use a secure random generator here for there is no cryptographically need in this case

daniel-y:
@yannaingtun could you please sign the CLA before merging this PR?

PR #2431, opened 2025-04-18
feat(metric): add cert metrics

woshigaopp:
comment to ping reviewers.*


SCNieh (changes requested):
There are several places that can be optimized to better fit with current coding style:
1. new metrics can be added directly into S3StreamKafkaMetricsManager, this is where we put all the AutoMQ-related metrics
2. you can refer to the usage of other gauge type metrics in S3StreamKafkaMetricsManager, which will typically have a corresponding supplier to for value retrieval in the callback of the gauge builder
3. the supplier can be defined in TelemetryManager during initialization and set into S3StreamKafkaMetricsManager

SCNieh (approved):

Issue #2364, opened 2025-03-18
Memory Leak in `CompactionUploader` during Shutting Down


Chillax-0v0:
```
2025-03-18 15:55:10.063 +0800 ERROR [io.netty.util.ResourceLeakDetector]  LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.AbstractByteBufAllocator.compositeDirectBuffer(AbstractByteBufAllocator.java:224)
	com.automq.stream.s3.ByteBufAlloc.compositeByteBuffer(ByteBufAlloc.java:128)
	com.automq.stream.s3.operator.ProxyWriter$ObjectWriter.<init>(ProxyWriter.java:127)
	com.automq.stream.s3.operator.ProxyWriter.<init>(ProxyWriter.java:33)
	com.automq.stream.s3.operator.ProxyWriter.<init>(ProxyWriter.java:48)
	com.automq.stream.s3.operator.AbstractObjectStorage.writer(AbstractObjectStorage.java:200)
	com.automq.stream.s3.compact.operator.DataBlockWriter.<init>(DataBlockWriter.java:59)
	com.automq.stream.s3.compact.CompactionUploader.lambda$writeStreamObject$5(CompactionUploader.java:116)
	java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1150)
	java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:482)
	java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	java.base/java.lang.Thread.run(Thread.java:840)
```

kiranraoboinapally:
assign

SCNieh:
@kiranraoboinapally  Hi, we appreciate your interest in this issue, however this is not a good first issue and may involve of a lot of core code changing, if this is your first time to AutoMQ, we recommend to start from other good first issues, thanks.

vishalmaurya850:
@Chillax-0v0 Assign this to me. I will resolve this issue.

PR #2440, opened 2025-04-22
fix(proxywriter): resolve memory leak by ensuring ByteBuf release

kiranraoboinapally:
small contrbution


SCNieh:
Hi, thanks for your commit, just some feedbacks on your change:
1. pls do not modify the code formation so we can see the real code difference
2. ObjectWriter#data is assigned during object initialization so it'll nevever be null when release() is called, thus the null check is unnccessary
3. the memory leak mention in #2364 is caused by missing release of ObjectWriter when CompactionUploader is interrupted because of process shutdown, the implementation of ObjectWriter is just fine
We'll close the PR since it's unrelated to the mentioned issue above, though thanks for your contribution anyway

Issue #2373, opened 2025-03-18
[Good First Issue] Add Support for Specifying Topic Prefix for Catch-up Reading in Performance Test


Chillax-0v0:
### Background
AutoMQ provides the [`automq-perf-test.sh`](https://www.automq.net/blog/how-to-perform-a-performance-test-on-automq) script to test performance under different scenarios. Currently, it supports testing "catch-up read" (cold read) scenarios by:
1. Sending a batch of messages.
2. Delaying consumption until messages have accumulated for a specified period.
3. Observing consumption throughput and its impact on producers.

But currently, each test requires spending significant time to accumulate messages, which slows down the test.

### Goal
Save testing time by allowing users to **pre-specify topics or topic groups (via a prefix)** for catch-up read testing.

### Required Changes
1. Add a parameter (e.g., `--catchup-topic-prefix`) to specify a topic prefix for catch-up read testing.  
2. If the prefix is provided, reuse existing topics matching the prefix instead of creating new ones.  
3. Ensure the script skips the message accumulation phase when using existing topics.

xiayu1118:
Hello, I have a few questions in the process of implementing this issue, and I hope to get answers. Regarding your optimization requirement: "By adding the -catchup-topic-prefix prefix, the script can reuse the existing topic", I found that the perfcommend class is the actual test tool. Do I need to modify the code of this class to verify the "-catchup-topic-prefix" prefix parameter, or can I meet the requirement by simply modifying the logic of automq-perf-test.sh?

Chillax-0v0:
> Hello, I have a few questions in the process of implementing this issue, and I hope to get answers. Regarding your optimization requirement: "By adding the -catchup-topic-prefix prefix, the script can reuse the existing topic", I found that the perfcommend class is the actual test tool. Do I need to modify the code of this class to verify the "-catchup-topic-prefix" prefix parameter, or can I meet the requirement by simply modifying the logic of automq-perf-test.sh?

Hello, @xiayu1118, the `automq-perf-test.sh` script primarily serves as an entry point that ultimately executes the `org.apache.kafka.tools.automq.PerfCommand#main` method. To implement this feature, you will need to modify the relevant Java code rather than just the shell script logic.

[...]

PR #2495, opened 2025-05-04
Add support for reusing topics in catch-up read testing

sakshamVerma08:
This PR addresses the issue of time-consuming performance tests when testing "catch-up read" scenarios. Currently, each test run requires spending significant time to accumulate messages, which slows down testing. This PR adds the ability to reuse existing topics with accumulated messages, eliminating the wait time.

Issue: #2373 

Added a new command-line parameter `--catchup-topic-prefix` that allows users to specify topics to reuse for catch-up read testing. When this parameter is provided and topics matching the prefix exist:

1. The script reuses these topics instead of creating new ones
2. The message accumulation phase is skipped
3. Consumers are reset to read from the beginning of the topics

This significantly speeds up testing by eliminating the wait time for message accumulation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,125 +1,54 @@
-# AutoMQ: A stateless Kafka® on S3, offering 10x cost savings and scaling in seconds.
+# Contributing to AutoMQ
 
-<div align="center">
-<p align="center">
-  📑&nbsp <a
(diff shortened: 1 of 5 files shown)


SCNieh:
@sakshamVerma08 Hi, thanks for your work, but your code seems incomplete and maybe you have missed some commits? The current one just added some variables and method that never been used, plus please do not modify the README.

PR #2547, opened 2025-05-14
Add JMX-based Prometheus Metrics Exporter for JVM Observability in AutoMQ

itzsandeepshrma:
This PR adds a JMX-based Prometheus metrics exporter to AutoMQ, exposing JVM metrics like memory usage, thread counts, and GC stats. A new JmxMetricsExporter.java class is introduced to start a Prometheus server on port 8080. ServerMain.java is updated to call JmxMetricsExporter.start(8080) at startup, and the pom.xml includes the simpleclient_hotspot dependency for Prometheus integration. This enables seamless monitoring with Prometheus and Grafana without affecting existing functionality.

### DIFF
--- a/automq-shell/pom.xml
+++ b/automq-shell/pom.xml
@@ -1,0 +1,5 @@
+<dependency>
+    <groupId>io.prometheus</groupId>
+    <artifactId>simpleclient_hotspot</artifactId>
+    <version>0.16.0</version>
+</dependency>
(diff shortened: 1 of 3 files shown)


SCNieh:
@itzsandeepshrma Hi, AutoMQ has already supported exposing metrics (including JVM metrics) as a Prometheus HTTP server, please check PrometheusMetricsExporter for more details.

itzsandeepshrma:
Thank you for the clarification. I reviewed the existing PrometheusMetricsExporter and see it already provides JVM metrics support. I will explore extending or improving the current implementation instead of adding duplicate functionality.

Could you please advise if there are specific areas or gaps in metrics coverage where contributions would be most valuable?

SCNieh:
@itzsandeepshrma thanks for your interest but there won't be much to do with the metrics in AutoMQ

itzsandeepshrma:
Okay thanks for your suggestion

PR #2573, opened 2025-05-17
feat: Added consumer percentage control for catch-up reading in perf test

### LINKED ISSUE #2372: [Good First Issue] Add Consumer Percentage Control for Catch-up Reading in Performance Test
### Background
AutoMQ provides the [`automq-perf-test.sh`](https://www.automq.com/blog/how-to-perform-a-performance-test-on-automq) script to test performance under different scenarios. Currently, it supports testing "catch-up read" (cold read) scenarios by:
1. Sending a batch of messages.
2. Delaying consumption until messages have accumulated for a specified period.
3. Observing consumption throughput and its impact on producers.

However, the current implementation starts **all consumers simultaneously**, which does not reflect real-world scenarios where only a **subset of topics** experience catch-up reads at the same time.

### Required Changes
Add a new **optional command-line argument** (e.g., `--consumers-during-catchup`) to the `automq-perf-test.sh` script to control the **num of consumers** activated during catch-up read scenarios.

sanks011:
This PR implements a new command-line parameter `--consumers-during-catchup` that allows users to control what percentage of consumer groups are activated during catch-up reads in performance testing.

This implementation adds:

A new `--consumers-during-catchup` parameter (0-100, default: 100%) to control the percentage of consumers activated during catch-up reads
Added the new parameter to (PerfConfig) class with proper validation
Enhanced (ConsumerService) with percentage-based consumer selection logic
Modified (PerfCommand) to utilize the new parameter during catch-up reads
Added documentation in the [automq-perf-test.sh] script

Resolves #2372

### DIFF
--- a/tools/src/main/java/org/apache/kafka/tools/automq/PerfCommand.java
+++ b/tools/src/main/java/org/apache/kafka/tools/automq/PerfCommand.java
@@ -169,7 +169,8 @@ public class PerfCommand implements AutoCloseable {
         Result result;
         if (config.backlogDurationSeconds > 0) {
             LOGGER.info("Pausing consumers for {} seconds to build up backlog...", config.backlogDurationSeconds);
-            consumerService.pause();            long backlogStart = System.currentTimeMillis();
+            consumerService.pause();
+            long backlogStart = System.currentTimeMillis();
             collectStats(Duration.ofSeconds(config.backlogDurationSeconds));
             long backlogEnd = System.nanoTime();
 
--- a/tools/src/main/java/org/apache/kafka/tools/automq/PerfCommand.java
+++ b/tools/src/main/java/org/apache/kafka/tools/automq/PerfCommand.java
@@ -171,12 +171,11 @@ public class PerfCommand implements AutoCloseable {
             LOGGER.info("Pausing consumers for {} seconds to build up backlog...", config.backlogDurationSeconds);
             consumerService.pause();
             long backlogStart = System.currentTimeMillis();
-            collectStats(Duration.ofSeconds(config.backlogDurationSeconds));
-            long backlogEnd = System.nanoTime();
+            collectStats(Duration.ofSeconds(config.backlogDurationSeconds));            long backlogEnd = System.nanoTime();
 
             LOGGER.info("Resetting consumer offsets and resuming...");
-            consumerService.resetOffset(backlogStart, TimeUnit.SECONDS.toMillis(config.groupStartDelaySeconds));
-            consumerService.resume();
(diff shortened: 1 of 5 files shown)


Chillax-0v0:
Hi, thank you for your contribution. I've reviewed the submitted code and it's well-structured and functional. However, there's an area that could be improved:

I noticed in your implementation that during catch-up read phase, you're scaling down the number of consumer groups per topic by the specified percentage. However, the parameter should actually control the number of topics being consumed during catch-up rather than affecting group counts per topic. This approach would better align with real-world scenarios.

To clarify with your example parameters: `--topics 10 --groups-per-topic 2 --consumers-during-catchup 50`

Current behavior:  
Each topic would have 2 * 50% = 1 consumer group during catch-up.

Desired behavior:  
Only 10 * 50% = 5 topics would be consumed during catch-up, each maintaining 2 consumer groups.

This adjustment would make the catch-up mechanism more logically consistent by controlling topic participation rather than group density. I'm happy to discuss this further or help clarify any implementation details.

sanks011:
@Chillax-0v0 can you check if this meets the requirement... I hope the implementation now correctly controls the number of topics with active consumers during catch-up, rather than scaling down the number of consumer groups per topic.

tools/src/main/java/org/apache/kafka/tools/automq/PerfCommand.java
```diff
@@ -176,7 +176,23 @@ private void run() {
...
+            List<String> resumeTopics = allTopicNames.subList(0, topicsToResume);
+            List<String> pauseTopics = allTopicNames.subList(topicsToResume, numTopics);
+
+            consumerService.pause(); // Pause all first
+            consumerService.resumeTopics(resumeTopics); // Resume only selected topics
+            LOGGER.info("Resuming consumers for topics: {} ({} out of {})", resumeTopics, topicsToResume, numTopics);
+            if (!pauseTopics.isEmpty()) {
+                consumerService.pauseTopics(pauseTopics);
+                LOGGER.info("Keeping consumers paused for topics: {} ({} out of {})", pauseTopics, pauseTopics.size(), numTopics);
+            }
```
Chillax-0v0, line 195: If I'm not mistaken, we have already paused all Consumers at L172. Is it necessary to pause them again at L189 and L193 here?

sanks011: > If I'm not mistaken, we have already paused all Consumers at L172. Is it necessary to pause them again at L189 and L193 here?

You're absolutely right! Thank you for catching that redundancy. 
I've updated the implementation to remove the unnecessary pause operations. The consumers are already paused at line 172 when we build up the backlog, so there's no need to pause them again before resuming selected topics.

Chillax-0v0 (approved):
LGTM

Issue #2343, opened 2025-03-12
[Enhancement] Enhance KVRecord


superhx:
### Who is this for and what problem do they have today?
AutoMQ team

### Why is solving this problem impactful?
Currently, the following types of data are stored in the KVRecord:
- Partition -> MetaStreamId: _kafka_$clusterId/$topicName/$partitionIndex -> $metaStreamId
- Node -> NodeMetadata: __automq_node/$nodeId -> $nodeMetadata.

KVs of different functions are isolated by the prefix of the key.

Expectations:
- Add namespace to KVRecord to more elegantly isolate (future) KVs used by different functions.
- Add epoch to KVRecord to avoid concurrent operations of PutKVRequest accidentally overwriting KVs.

yyyCode:
I couldn't find KVRecord in the AutoMQ source code, is it part of AutoMQ or its dependencies?

AminMortezaie:
Hi team 👋,

I’d like to take this on if it's still unassigned.

I’ll investigate the reassignment path and ensure it gracefully handles the case where a partition has no segments (e.g., by skipping or initializing default state).

I’ll also add a test to cover this edge case. Let me know if there are any preferred conventions or context I should keep in mind!

Best,  
Amin

AminMortezaie:
I've added my merge request.

1sonofqiu:
@AminMortezaie Thanks for your contribution! I'll take a look at it within a few days.

PR #2581, opened 2025-05-20
feat: add namespace and epoch for enhancement.

AminMortezaie:
# Add Namespace and Epoch Support to KV Store

## Overview
This PR introduces namespace and epoch support to AutoMQ's KV store system through a composable design pattern, allowing for backward compatibility while providing new functionality for applications that need namespace isolation and epoch-based versioning.

## Key Changes
1. Added a new `NamespacedControllerKVClient` that extends the base `ControllerKVClient`
2. Introduced a feature flag `namespacedKVEnabled` in `Config` for opt-in functionality
3. Modified `DefaultS3Client` to support both namespaced and non-namespaced modes

## Implementation Details
- **Namespace Support**: Keys are prefixed with their namespace (e.g., "namespace:key") for isolation
- **Epoch Management**: Each namespace maintains its own epoch counter for versioning
- **Backward Compatibility**: 
  - Default behavior remains unchanged (namespacedKVEnabled = false)
  - Existing applications continue to work without modification
  - New applications can opt-in to namespaced functionality

## Usage
To enable namespaced KV operations:
```java
Config config = new Config()
    .namespacedKVEnabled(true);
```

When enabled, applications can:
- Use namespace isolation for keys
- Maintain epoch versioning per namespace
- Prevent cross-namespace conflicts
- Maintain data isolation between different parts of the system

## Migration Path
1. Deploy the changes with namespacedKVEnabled = false (default)
2. Applications can gradually opt-in to namespaced functionality
3. No immediate action required for existing applications

## Performance Impact
- Minimal overhead for non-namespaced operations (no changes to existing path)
- Small additional overhead for namespaced operations:
  - Key prefix addition
  - Epoch checks
  - Additional epoch storage per namespace

## Future Considerations
1. Namespace cleanup utilities
2. Epoch management tools
3. Namespace statistics and monitoring
4. Cross-namespace operations (if needed)

## Related Issue
- [Issue #2343 ] [Enhancement] Enhance KVRecord

### DIFF
--- a/core/src/main/scala/kafka/log/stream/s3/ControllerKVClient.java
+++ b/core/src/main/scala/kafka/log/stream/s3/ControllerKVClient.java
@@ -55,7 +55,7 @@ import java.util.concurrent.CompletableFuture;
 public class ControllerKVClient implements KVClient {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ControllerKVClient.class);
-    private final ControllerRequestSender requestSender;
+    protected final ControllerRequestSender requestSender;
 
     public ControllerKVClient(ControllerRequestSender requestSender) {
         this.requestSender = requestSender;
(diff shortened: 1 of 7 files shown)


superhx:
@AminMortezaie  We sincerely appreciate your contributions. However, it appears that there might be a misunderstanding regarding the requirement. Instead of manipulating the key field, what we actually need is to extend the KVRecord by adding two new fields: namespace and epoch.

- With namespace field, we can fast filter the KVs in certain namespace instead of matching them by prefix.

AminMortezaie:
Great, I will change to the real requirements.

PR #2627, opened 2025-06-16
fix(docker): update MinIO command from config host to alias set

Gezi-lzq:
https://github.com/minio/mc/issues/5206

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -46,7 +46,7 @@ services:
         condition: service_healthy
     entrypoint: >
       /bin/sh -c "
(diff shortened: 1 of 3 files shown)


docker/docker-compose.yaml
```diff
@@ -46,7 +46,7 @@ services:
         condition: service_healthy
     entrypoint: >
       /bin/sh -c "
-      until (/usr/bin/mc config host add minio http://minio:9000 minioadmin minioadmin) do echo '...waiting...' && sleep 1; done;
+      until (/usr/bin/mc alias set minio http://minio:9000 minioadmin minioadmin) do echo '...waiting...' && sleep 1; done;
```
superhx, line 49: We should specify the minio version to prevent it break again

PR #2659, opened 2025-07-08
fix(s3stream): avoid StreamMetadataManager add callback when retry

lifepuzzlefun:
comment to ping reviewers.*

### DIFF
--- a/core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
+++ b/core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
@@ -130,12 +130,12 @@ public class StreamMetadataManager implements InRangeObjectsFetcher, MetadataPub
     public CompletableFuture<InRangeObjects> fetch(long streamId, long startOffset, long endOffset, int limit) {
         // TODO: cache the object list for next search
         CompletableFuture<InRangeObjects> cf = new CompletableFuture<>();
-        exec(() -> fetch0(cf, streamId, startOffset, endOffset, limit), cf, LOGGER, "fetchObjects");
+        exec(() -> fetch0(cf, streamId, startOffset, endOffset, limit, false), cf, LOGGER, "fetchObjects");
         return cf;
     }
 
     private void fetch0(CompletableFuture<InRangeObjects> cf, long streamId,
-        long startOffset, long endOffset, int limit) {
+        long startOffset, long endOffset, int limit, boolean retryFetch) {
         Image image = getImage();
         try {
             final S3StreamsMetadataImage streamsImage = image.streamsMetadata();
@@ -171,9 +171,11 @@ public class StreamMetadataManager implements InRangeObjectsFetcher, MetadataPub
                     streamId, startOffset, endOffset, limit, rst.objects().size(), rst.endOffset());
 
                 CompletableFuture<Void> pendingCf = pendingFetch();
-                pendingCf.thenAccept(nil -> fetch0(cf, streamId, startOffset, endOffset, limit));
-                cf.whenComplete((r, ex) ->
-                    LOGGER.info("[FetchObjects],[COMPLETE_PENDING],streamId={} startOffset={} endOffset={} limit={}", streamId, startOffset, endOffset, limit));
+                pendingCf.thenAccept(nil -> fetch0(cf, streamId, startOffset, endOffset, limit, true));
+                if (!retryFetch) {
+                    cf.whenComplete((r, ex) ->
+                        LOGGER.info("[FetchObjects],[COMPLETE_PENDING],streamId={} startOffset={} endOffset={} limit={}", streamId, startOffset, endOffset, limit));
+                }
             }).exceptionally(ex -> {
                 cf.completeExceptionally(ex);
                 return null;


lifepuzzlefun:
image

image


maybe the metadataImage have problem. this fix only resolve too much callback addon the fetch0 cf.

Gezi-lzq:
> maybe the metadataImage have problem. this fix only resolve too much callback addon the fetch0 cf.

Thanks for the detailed report. We've recently encountered a very similar issue, and the symptoms and root cause appear to be fundamentally the same.

Here's a summary of what we've observed:

The primary symptom is a memory leak, observable as a slow but steady increase in memory usage. When we take a heap dump, we can see that the majority of the retained memory consists of a large number of `CompletableFuture` objects under the `StreamMetadataManager`.

The immediate cause of this is a request that triggers continuous asynchronous recursion in `fetch0`, generating new `CompletableFuture` instances. These futures have mutual references, which prevents them from being garbage collected.

The ultimate root cause, however, lies in why the data fetch fails in the first place. After investigation, we believe the main reason is a concurrency issue during the `compaction` phase. A concurrent access to a non-thread-safe `Map` results in `Map#get` returning `null` (this is particularly likely to happen if a `Map#put` triggers a resize at the same moment). Consequently, the `Compaction` process incorrectly assumes the stream does not exist and discards its corresponding data.

We have already submitted a pull request to address this issue: [https://github.com/AutoMQ/automq/pull/2653](https://github.com/AutoMQ/automq/pull/2653).

core/src/main/scala/kafka/log/stream/s3/metadata/StreamMetadataManager.java
```diff
@@ -171,9 +171,11 @@ private void fetch0(CompletableFuture<InRangeObjects> cf, long streamId,
                     streamId, startOffset, endOffset, limit, rst.objects().size(), rst.endOffset());
 
                 CompletableFuture<Void> pendingCf = pendingFetch();
-                pendingCf.thenAccept(nil -> fetch0(cf, streamId, startOffset, endOffset, limit));
-                cf.whenComplete((r, ex) ->
-                    LOGGER.info("[FetchObjects],[COMPLETE_PENDING],streamId={} startOffset={} endOffset={} limit={}", streamId, startOffset, endOffset, limit));
+                pendingCf.thenAccept(nil -> fetch0(cf, streamId, startOffset, endOffset, limit, true));
```
Gezi-lzq, line 174: My current thinking is that the primary cause of the memory leak is likely the callback created for `pendingCf.thenAccept(...)`. I believe the closure for this callback captures the original cf and all the arguments for the `fetch0` method. This would hold a reference to the new cf, preventing it from being garbage collected by the gc.

lifepuzzlefun: yeah, the pendingCf can be completed.

so for some reason the fetch0 is always retry.
each time retry the cf.whenComplete execute once. so the callback make the memory leak.

we only need add complete once to do log

lifepuzzlefun:
> > maybe the metadataImage have problem. this fix only resolve too much callback addon the fetch0 cf.
> 
> Thanks for the detailed report. We've recently encountered a very similar issue, and the symptoms and root cause appear to be fundamentally the same.
> 
> Here's a summary of what we've observed:
> 
> The primary symptom is a memory leak, observable as a slow but steady increase in memory usage. When we take a heap dump, we can see that the majority of the retained memory consists of a large number of `CompletableFuture` objects under the `StreamMetadataManager`.
> 
> The immediate cause of this is a request that triggers continuous asynchronous recursion in `fetch0`, generating new `CompletableFuture` instances. These futures have mutual references, which prevents them from being garbage collected.
> 
> The ultimate root cause, however, lies in why the data fetch fails in the first place. After investigation, we believe the main reason is a concurrency issue during the `compaction` phase. A concurrent access to a non-thread-safe `Map` results in `Map#get` returning `null` (this is particularly likely to happen if a `Map#put` triggers a resize at the same moment). Consequently, the `Compaction` process incorrectly assumes the stream does not exist and discards its corresponding data.
> 
> We have already submitted a pull request to address this issue: #2653.

I'll try to use this patch in my env. reply if any progress.

superhx (approved):

lifepuzzlefun:
just share what i found:

```
ERROR [streamId=7256 epoch=77] stream fetch [6058002581383, 606014 1241704) 65536 fail (com.automq.stream.s3.S3Stream)
```

this is the always fail metadataImage param

```
# streamMetadata
startOffset = long 6058002581383


# SO
objectId = long 3189001535
startOffset = long 6060151248961
endOffset = long   6060152230213


# RangeMetadata
range0
nodeId = int 1044
startOffset = long 6058002581383
endOffset =   long 6060152230213
rangeIndex = int 39

range1
nodeId = 1017
startOffset = long 6060152230213
endOffset = long 6060152230213

range2
startOffset = long 6060152230213
endOffset =   long 6060152230213
```

our environment is luck that the stream only have those metadataImage

only one SO and 3 Range

so the query metadata should find in range0 to get the whole metadata, right? 

and the query should query from NodeRangeIndexCache.

but it return null, I think maybe some bug make the stream is lost in the NodeRangeIndexCache

 
cc @superhx @Gezi-lzq

lifepuzzlefun:
image
image
image

superhx:
@lifepuzzlefun 
> so the query metadata should find in range0 to get the whole metadata, right?

yes

> and the query should query from NodeRangeIndexCache.

The NodeRangeIndexCache is a sparse index, it's just for query acceleration.

The fetch failure will be fixed after the data expired.

PR #2665, opened 2025-07-11
feat(s3stream/block-wal): complete appends sequentially

Chillax-0v0:


s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/SlidingWindowService.java
```diff
@@ -529,13 +569,25 @@ public void run() {
...
+        }
 
+        private void wroteBlock(Block block) {
             final long startTime = System.nanoTime();
+            WroteBlockResult wroteBlockResult = SlidingWindowService.this.wroteBlock(block);
             // Update the start offset of the sliding window after finishing writing the record.
-            windowCoreData.updateWindowStartOffset(wroteBlock(block));
+            windowCoreData.updateWindowStartOffset(wroteBlockResult.unflushedOffset);
+            // Notify the futures of blocks that have been written.
+            for (Block writtenBlock : wroteBlockResult.writtenBlocks) {
```
superhx: Sequentially callback should guard by lock

s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/BlockWALService.java
```diff
@@ -885,47 +858,21 @@ private boolean tryReadNextRecord() throws RuntimeIOException {
...
                     if (reportError) {
                         next = new InvalidRecoverResult(nextRecoverOffset, e.getMessage());
-                    }
-                    nextRecoverOffset = e.getJumpNextRecoverOffset();
-                    if (reportError) {
                         return true;
                     }
+                    if (WALUtil.isAligned(nextRecoverOffset) && !skip) {
+                        LOGGER.info("meet the first invalid offset during recovery. offset: {}, detail: '{}'", nextRecoverOffset, e.getMessage());
+                        return false;
```
allenzhu: SlidingWindowService.sealAndNewBlockLocked() :
```
if ((recordSectionCapacity - startOffset % recordSectionCapacity) < minSize) {
    startOffset = startOffset + recordSectionCapacity - startOffset % recordSectionCapacity;
}
```
As shown in the code：
When the remaining space in the sliding window is insufficient to accommodate the message size, the startOffset moves to the beginning of the recordSectionCapacity. If the message size exceeds BLOCK_SIZE, this creates holes larger than BLOCK_SIZE in the storage. As a result: Recovered data becomes non-continuous, subsequent data risks being truncated.

Chillax-0v0: Good catch! We'll fix this in the next PR.

Chillax-0v0: Done in #2673.

PR #2673, opened 2025-07-15
feat(s3stream/wal): write a padding record when no space at the end of device

Chillax-0v0:


s3stream/src/main/java/com/automq/stream/s3/wal/impl/block/SlidingWindowService.java
```diff
@@ -418,8 +456,8 @@ private WroteBlockResult wroteBlock(Block wroteBlock) {
     private WroteBlockResult wroteBlockLocked(Block wroteBlock) {
         wroteBlock.markWritten();
 
+        assert writingBlocks.contains(wroteBlock);
         List<Block> writtenBlocks = removeWrittenBlocksLocked();
-        assert !writtenBlocks.isEmpty();
```
allenzhu, line 422: writtenBlocks may be empty

Chillax-0v0: Fixed.

PR #2691, opened 2025-07-22
fix: update comment to reflect cluster setup in docker-compose-cluste…

Napat:
fix: update comment to reflect cluster setup in docker-compose-cluster.yaml


Gezi-lzq:
Nice catch! LTGM

PR #2708, opened 2025-07-28
perf(s3stream): prewarm MetaStream before stream ready to open

lifepuzzlefun:
comment to ping reviewers.*


lifepuzzlefun:
awaitStreamReadyForOpen often cost 500ms+, but we can trigger prewarm the first need 64KB for metastream to reduce the read time.

superhx:
> awaitStreamReadyForOpen often cost 500ms+, but we can trigger prewarm the first need 64KB for metastream to reduce the read time.

@lifepuzzlefun  The 500ms+ usually is the time cost of uploading WAL data to S3. Even if we pre-warm MetaStream, I think the MetaStream still needs to load more data from S3.

Issue #2718, opened 2025-07-29
capacity mismatch between relativeoffset and count constraints in elasticlogsegment


Gezi-lzq:
### Summary

A discrepancy exists between two `int`-based constraints governing the maximum capacity of an `ElasticLogSegment`. The segment's roll-over check is based on the `relativeOffset` limit inherited from Kafka, while the underlying `ElasticLogFileRecords` implementation is limited by a record `count`. This can lead to an integer overflow when a segment is almost full.

### Core Discrepancy

1.  **Relative Offset Constraint**: Based on Kafka's `OffsetIndex`, this allows a segment's relative offset (`offset - baseOffset`) to reach `Integer.MAX_VALUE`.

2.  **Record Count Constraint**: Required by `ElasticLogFileRecords`, the number of records in a batch (`count`) must fit within an `int`. The calculation `count = lastOffset - nextOffset` means the maximum count is `Integer.MAX_VALUE`.

The `relativeOffset` constraint is off by one compared to the `count` constraint. A write that results in a `relativeOffset` of `Integer.MAX_VALUE` would produce a `count` of `Integer.MAX_VALUE + 1`, causing an overflow.

### Areas for Adjustment

To resolve this, all logic related to segment capacity must be aligned with the stricter `count`-based constraint. The primary areas for review are:

1.  **Segment Rolling Logic (`ElasticLogSegment.shouldRoll` & `canConvertToRelativeOffset`)**: The capacity check must be tightened to prevent a write where the resulting `count` would exceed `Integer.MAX_VALUE`.

2.  **Log Cleaner Segment Grouping (`LogCleaner.groupSegmentsBySize`)**: The logic that merges segments checks if the new offset range exceeds `Integer.MAX_VALUE`. This should be reviewed to ensure it is consistent with the `count` constraint.

The goal is to unify the definition of segment capacity around the `count` limit imposed by the storage implementation.

superhx:
Could we rename the lastOffset of ElasticLogFileRecords to endOffset?

PR #2720, opened 2025-07-29
fix(log): Prevent potential offset overflow in ElasticLogSegment

Gezi-lzq:
This commit addresses an issue where a log segment could accommodate more than Integer.MAX_VALUE records, leading to a potential integer overflow when calculating relative offsets.

The root cause was that the check `offset - baseOffset <= Integer.MAX_VALUE` allowed a relative offset to be exactly `Integer.MAX_VALUE`. Since offsets are 0-based, this allows for `Integer.MAX_VALUE + 1` records, which cannot be represented by a standard Integer.

This fix implements the following changes:
1.  In `ElasticLogSegment`, the offset validation is changed from `<=` to `< Integer.MAX_VALUE` to ensure the relative offset strictly fits within an Integer's bounds.
2.  In `LogCleaner`, a new segment grouping method `groupSegmentsBySizeV2` is introduced for `ElasticUnifiedLog`. This method uses the same stricter offset check to prevent incorrectly grouping segments that would exceed the offset limit.
3.  The corresponding unit tests in `LogCleanerTest` have been updated to reflect these new boundaries and validate the fix.

Fixes: #2718

### DIFF
--- a/core/src/main/scala/kafka/log/LogCleaner.scala
+++ b/core/src/main/scala/kafka/log/LogCleaner.scala
@@ -1002,7 +1002,8 @@ private[log] class Cleaner(val id: Int,
             timeIndexSize + segs.head.timeIndex.sizeInBytes <= maxIndexSize &&
             //if first segment size is 0, we don't need to do the index offset range check.
             //this will avoid empty log left every 2^31 message.
-            (segs.head.size == 0 || isOffsetRangeValid(group))) {
+            (segs.head.size == 0 ||
+              isOffsetRangeValid(group))) {
         group = segs.head :: group
         logSize += segs.head.size
         indexSize += offsetIndexSize(segs.head)
--- a/core/src/main/scala/kafka/log/LogCleaner.scala
+++ b/core/src/main/scala/kafka/log/LogCleaner.scala
@@ -1003,7 +1003,9 @@ private[log] class Cleaner(val id: Int,
             //if first segment size is 0, we don't need to do the index offset range check.
             //this will avoid empty log left every 2^31 message.
             (segs.head.size == 0 ||
+              // AutoMQ inject start
               isOffsetRangeValid(group))) {
+              // AutoMQ inject end
         group = segs.head :: group
         logSize += segs.head.size
         indexSize += offsetIndexSize(segs.head)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -984,13 +985,19 @@ private[log] class Cleaner(val id: Int,
...
+        timeIndexSize + segs.head.timeIndex.sizeInBytes <= maxIndexSize &&
+        //if first segment size is 0, we don't need to do the index offset range check.
+        //this will avoid empty log left every 2^31 message.
+        (segs.head.size == 0 || {
+          val offsetRange = lastOffsetForFirstSegment(segs, firstUncleanableOffset) - group.last.baseOffset
+          // For ElasticLogSegment, use a stricter offset range check (`< Int.MaxValue`) to prevent a potential overflow
+          // issue as described in https://github.com/AutoMQ/automq/issues/2717.
+          // For other segment types, the original less-strict check (`<= Int.MaxValue`) is retained.
+          if (group.last.isInstanceOf[ElasticLogSegment]) offsetRange < Int.MaxValue
+          else offsetRange <= Int.MaxValue
```
superhx: - Do we need to make the difference for ElasticLogSegment and LogSegment?
- Keep the old format to avoid future merge conflict

superhx:
Add S3Stream#append RecordBatch.count check

core/src/main/scala/kafka/log/LogCleaner.scala
Gezi-lzq: The purpose of this modification is to prevent the segment from exceeding the constraint due to cleaning.
Initially, `groupSegmentsBySize` uses `lastOffsetForFirstSegment(segs, firstUncleanableOffset) - group.last.baseOffset <= Int.MaxValue` to ensure that the offset range of segments added to the same group does not exceed Int.MaxValue.
However, ElasticLogSegment has a new constraint that requires ensuring `count <= Int.MaxValue` in a Segment, so the condition here needs to be changed: `lastOffsetForFirstSegment(segs, firstUncleanableOffset) - group.last.baseOffset <= Int.MaxValue`

Since this constraint only exists on ElasticLogSegment, a distinction is made between ElasticLogSegment and LogSegment.

Issue #2723, opened 2025-07-29
Test suite failures on Java 21


vivekchavan14:
Environment:
•  OS: Ubuntu 24.04.2 LTS
•  Java: OpenJDK 21.0.8 (Ubuntu 21.0.8+9-Ubuntu-0ubuntu124.04.1)
•  Gradle: 8.8
•  AutoMQ Commit: 014d2d2861 (latest main branch)
•  Command: ./gradlew build

Issue Description:
The test suite fails with 132 total test failures when building on Java 21. The build succeeds when tests are skipped (./gradlew build -x test), indicating this is purely a test environment issue.

Failure Summary:
•  Metadata module: 131/931 tests failing (85% success rate)
•  Connect API module: 1/313 tests failing (99% success rate)

Major Failing Test Classes:
•  ClusterControlManagerTest (21/22 tests failing)
•  ReplicationControlManagerTest (64/66 tests failing) 
•  QuorumControllerTest (24/34 tests failing)
•  ProducerIdControlManagerTest (all 5 tests failing)
•  QuorumControllerMetricsIntegrationTest (all 5 tests failing)
•  ValuesTest.shouldParseStringListWithMultipleElementTypes() (connect-api)

Sample Error:
```bash
org.opentest4j.AssertionFailedError: expected: <MAP> but was: <ARRAY>
    at org.apache.kafka.connect.data.ValuesTest.assertParseStringArrayWithNoSchema(ValuesTest.java:478)
```

Workaround:
Build succeeds with
```bash
./gradlew build -x test
```

superhx:
Hi, The original Apache Kafka test cases have not been fully adapted yet. Therefore, it is necessary to run the test cases through ./gradlew build -x test.

Welcome to participate in the test fix

vivekchavan14:
Thanks for the clarification!
I'll proceed with 
```bash
./gradlew build -x test,
```

I'll try to fix test-cases

Bhuvannnn:
@superhx , the report is 132 failures on Java 21. 131 of those are metadata/KRaft controller tests. The other one is connect-api `ValuesTest.shouldParseStringListWithMultipleElementTypes` (expected MAP, got ARRAY).

Would a first PR that only does that ValuesTest case be okay or do you want the metadata failures in the same change?

superhx:
[...]

PR #2724, opened 2025-07-29
refactor: clarify callback responsibility and remove redundant topic …

### LINKED ISSUE #2721: refactor: clarify `callback` responsibility and remove redundant topic cleanup logic
### 1. Background: The Role of `asyncApplyDelta` and its `callback`

In `BrokerMetadataPublisher#handleTopicsDelta`, core partition state changes are asynchronously applied via `ElasticReplicaManager#asyncApplyDelta`.

This method includes a `callback` mechanism designed for a **fast response to single-partition state changes**. For instance, after a partition becomes a leader (`makeLeader`) on the local broker, the `callback` is immediately invoked. This allows modules like `GroupCoordinator` to react quickly to the state change for that specific partition without waiting for all other operations in the `Delta` to complete, which is crucial for timely state synchronization.

### 2. Evolution of the Issue and Logic Redundancy

The logic for cleaning up consumer group offsets has evolved through two main iterations, leading to the current redundancy.

*   **Phase 1 (PR #1264):** The initial fix extended the `callback`'s role to address cleanup after topic deletion. It added logic to iterate over all partitions of a deleted topic and invoke the `callback` for them if any replicas were found in `localChanges.deletes`.

*   **Phase 2 (PR #2626):** It was later found that the fix in PR #1264 was incomplete. If a broker held no replicas for the deleted topic, `localChanges.deletes` would be empty, and the cleanup logic would never trigger on that node. To resolve this, PR #2626 introduced a new approach: the cleanup logic was removed from the `callback` and moved to a new method, `notifyGroupCoordinatorOfDeletedPartitions`. This method is called after the `asyncApplyDelta` task completes (`whenComplete` stage) and operates on the global `topicsDelta.deletedTopicIds()`, ensuring reliable execution on all brokers.

### 3. Current State and Refactoring Proposal

The `callback` in `asyncApplyDelta` lacks a clear definition of responsibility at the interface level. This has led to it being used in two different scenarios: one for fast, single-partition state changes, and another for bulk, topic-level cleanup.

With the merge of PR #2626, the old cleanup logic (based on the `callback`) has become redundant, as its functionality is now superseded by the new implementation.

We propose removing the redundant logic to eliminate code redundancy and standardize the `callback`'s responsibility.

vivekchavan14:
…cleanup logic

Fixes #2721

- Remove redundant topic deletion cleanup logic from asyncApplyDelta callback
- The callback now focuses only on fast, single-partition state changes
- Topic-level cleanup is now handled exclusively by notifyGroupCoordinatorOfDeletedPartitions
- This eliminates code redundancy and clarifies the callback's responsibility

The redundant logic (lines 317-331) was superseded by the more reliable notifyGroupCoordinatorOfDeletedPartitions method introduced in PR #2626.

comment to ping reviewers.*

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
@@ -34,6 +34,7 @@ import com.automq.stream.s3.model.StreamRecordBatch;
 import com.automq.stream.s3.objects.ObjectManager;
 import com.automq.stream.utils.FutureUtil;
 import com.automq.stream.utils.LogSuppressor;
+import com.automq.stream.utils.Time;
 import com.automq.stream.utils.threads.EventLoop;
 import com.google.common.annotations.VisibleForTesting;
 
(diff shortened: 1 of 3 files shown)


Gezi-lzq:
Thank you for your contribution.
1. `connect/api/src/test/java/org/apache/kafka/connect/data/ValuesTest.java` seems to be irrelevant to the goal of this PR
2. The modification of `core/src/main/scala/kafka/server/metadata/BrokerMetadataPublisher.scala` does not seem to satisfy this problem and will cause errors.

vivekchavan14:
Yup, 

1. connect/api/src/test/java/org/apache/kafka/connect/data/ValuesTest.java, was trying to fix Kafka test-cases 
issue - https://github.com/AutoMQ/automq/issues/2723#issuecomment-3136049123, I mistakely pushed it here.
2. I'll go through BrokerMetaDataPublisher again,

PR #2814, opened 2025-09-05
feat(table): integrate RecordProcessor architecture and enhance error tolerance

Gezi-lzq:


core/src/main/java/kafka/automq/table/worker/IcebergWriter.java
```diff
@@ -93,22 +102,21 @@ public void write(int partition, org.apache.kafka.common.record.Record kafkaReco
...
-                long recordOffset = kafkaRecord.offset();
-                if (recordOffset < v.end) {
-                    throw new IllegalArgumentException(String.format("The record offset[%s] is less than the end[%s]", recordOffset, v.end));
-                }
-                v.end = recordOffset + 1;
-                return v;
-            });
-        } catch (InvalidDataException e) {
-            INVALID_DATA_LOGGER.warn("[INVALID_DATA],{}", this, e);
+            if (write0(partition, kafkaRecord)) {
```
superhx, line 105: We should record the offsetRangeMap even write0 return false

PR #2876, opened 2025-09-22
feat(metrics): integrate new metrics for table delay, fields per second, and event loop busy ratio

Gezi-lzq:


core/src/main/java/kafka/automq/table/worker/IcebergWriter.java
```diff
@@ -439,6 +456,61 @@ private boolean flush() throws IOException {
...
+        for (DataFile file : writeResult.dataFiles()) {
+            bytes += Math.max(file.fileSizeInBytes(), 0L);
+        }
+        for (DeleteFile file : writeResult.deleteFiles()) {
+            bytes += Math.max(file.fileSizeInBytes(), 0L);
+        }
+        return bytes;
+    }
+
+    private void waitForNetworkPermit() throws IOException {
```
superhx: There are duplicated codes. Maybe we could add a blocking consume to NetworkLimiter?

PR #2892, opened 2025-09-24
test(iceberg): ensure iceberg rest catalog supports custom headers passing th…

rmannibucau:
…rough headers to iceberg library

comment to ping reviewers.*

### DIFF
--- a/core/src/test/java/kafka/automq/table/CatalogFactoryTest.java
+++ b/core/src/test/java/kafka/automq/table/CatalogFactoryTest.java
@@ -1,10 +1,13 @@
 package kafka.automq.table;
 
-import com.sun.net.httpserver.HttpServer;
 import kafka.server.KafkaConfig;
-import org.apache.iceberg.rest.RESTCatalog;
+
 import org.apache.kafka.raft.QuorumConfig;
 import org.apache.kafka.server.config.KRaftConfigs;
+
+import com.sun.net.httpserver.HttpServer;
+
+import org.apache.iceberg.rest.RESTCatalog;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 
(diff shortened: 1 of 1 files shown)


Gezi-lzq:
Nice coverage here with the REST catalog setup. To make it bulletproof, could we also assert that the custom header (x-custom: my-x) is present on the request? Right now the handler only checks the URL, so the test would still pass if the header weren’t forwarded.

core/src/test/java/kafka/automq/table/CatalogFactoryTest.java
```diff
@@ -0,0 +1,89 @@
...
+                "automq.table.topic.catalog.type", "rest",
+                "automq.table.topic.catalog.uri", "http://localhost:" + catalogBackend.getAddress().getPort(),
+                "automq.table.topic.catalog.header.x-custom", "my-x", // Apache Polaris needs a tenant header for ex
+                "automq.table.topic.catalog.io-impl", "org.apache.iceberg.aws.s3.S3FileIO",
+                // automq specific/enforced (not standard catalog passthrough)
+                "s3.data.buckets", "0@s3://my_bucket?region=us-east-1&endpoint=http://localhost:12345&pathStyle=true"
+            )));
+            final var catalog = new CatalogFactory.Builder(config).build();
+            assertInstanceOf(RESTCatalog.class, catalog).close();
+            assertEquals(List.of("GET /v1/config?warehouse=s3://my_bucket/iceberg"), requests);
```
Gezi-lzq: Nice coverage here with the REST catalog setup. To make it bulletproof, could we also assert that the custom header (x-custom: my-x) is present on the request? Right now the handler only checks the URL, so the test would still pass if the header weren’t forwarded.

rmannibucau: good catch (was the original intent), updated

Gezi-lzq (approved):

Gezi-lzq:
LGTM

PR #2914, opened 2025-09-29
fix(log): elastic log supports truncate

superhx:


Gezi-lzq:
cherry-pick: https://github.com/AutoMQ/automq/pull/2913

PR #2979, opened 2025-10-31
fix: [iceberg] support avro union types in custom mapping

rmannibucau:
comment to ping reviewers.*

### DIFF
--- a/core/src/main/java/kafka/automq/table/binder/AbstractTypeAdapter.java
+++ b/core/src/main/java/kafka/automq/table/binder/AbstractTypeAdapter.java
@@ -57,38 +57,42 @@ public abstract class AbstractTypeAdapter<S> implements TypeAdapter<S> {
 
         switch (targetType.typeId()) {
             case BOOLEAN:
-                return convertBoolean(sourceValue, sourceSchema, targetType);
+                return convertBoolean(sourceValue, unwrapSchema(sourceSchema), targetType);
             case INTEGER:
-                return convertInteger(sourceValue, sourceSchema, targetType);
+                return convertInteger(sourceValue, unwrapSchema(sourceSchema), targetType);
             case LONG:
-                return convertLong(sourceValue, sourceSchema, targetType);
+                return convertLong(sourceValue, unwrapSchema(sourceSchema), targetType);
             case FLOAT:
-                return convertFloat(sourceValue, sourceSchema, targetType);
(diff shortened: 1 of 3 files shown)


Gezi-lzq:
1. https://github.com/AutoMQ/automq/blob/18340f2b950ee2f9be21ae3616a99ca740cd2452/core/src/test/java/kafka/automq/table/binder/AvroRecordBinderTest.java#L1160
2. https://github.com/AutoMQ/automq/blob/18340f2b950ee2f9be21ae3616a99ca740cd2452/core/src/test/java/kafka/automq/table/binder/AvroRecordBinderTest.java#L1056

Hello @rmannibucau , I would like to understand the situation this PR aims to address, and what are the main differences compared to that test case.

rmannibucau:
@Gezi-lzq you can remove the fix in the adapter and see how it fails. The diff is that existing test uses a struct which has a particular handling and my test a "primitive" which was not unwrapping the union and therefore not handling the logical type properly

Gezi-lzq:
Good catch! @rmannibucau  It seems we indeed missed handling a scenario for the timestamp type here.
When a timestamp type is part of a union type, `sourceSchema.getLogicalType()` fails to return a value at this location, leading to incorrect processing of the `long` variable.

https://github.com/AutoMQ/automq/blob/18340f2b950ee2f9be21ae3616a99ca740cd2452/core/src/main/java/kafka/automq/table/binder/AvroValueAdapter.java#L97-L100


I believe the correct fix would be to add the condition `Type.TypeID.TIMESTAMP.equals(icebergType.typeId())` to the logic here: https://github.com/AutoMQ/automq/blob/18340f2b950ee2f9be21ae3616a99ca740cd2452/core/src/main/java/kafka/automq/table/binder/RecordBinder.java#L159-L161

rmannibucau:
@Gezi-lzq sounds good, do you take it over or do you want me to update the pr next week? Happy to close this pr while a new release works for unions ;)? (Im afk until next week)

Gezi-lzq:
> @Gezi-lzq sounds good, do you take it over or do you want me to update the pr next week? Happy to close this pr while a new release works for unions ;)? (Im afk until next week)

Appreciate you pointing this out! Your work here definitely shows we need better test coverage for Union types.
I'm taking ownership of this to strengthen the testing groundwork first. Thanks again for the valuable input!
I'll ping you next week once I commit the changes, so you can easily review them. @rmannibucau

PR #2984, opened 2025-11-03
perf(s3stream): compute compaction delay using min timestamp instead of sorting

JasirVoriya:
**Issue Title**
Performance: avoid sorting when computing `compactionDelayTime`; use min timestamp

**Description**
In `CompactionManager.start()`’s periodic monitor task, `compactionDelayTime` is computed by sorting server objects by `committedTimestamp` and then taking the first element. This sorting is unnecessary because we only need the minimum timestamp. Sorting costs O(n log n) time; finding the minimum is O(n). With large object lists, the sort introduces avoidable CPU overhead and potential GC pressure.

**Affected Code**
- Path: `s3stream/src/main/java/com/automq/stream/s3/compact/CompactionManager.java`
- Method: `start()`
- Current logic:
```
data.sort(Comparator.comparingLong(S3ObjectMetadata::committedTimestamp));
this.compactionDelayTime = System.currentTimeMillis() - data.get(0).committedTimestamp();
```

**Benefits**
- Reduce time complexity from O(n log n) to O(n).
- Lower CPU usage and reduce GC pressure from sorting.
- Improve efficiency of the periodic monitoring task in high-object-count scenarios.

**Behavior**
- Semantics unchanged: the delay is still based on the earliest `committedTimestamp`.
- Edge cases unchanged: when `data` is null/empty or an exception occurs, delay is set to `0`.

**Validation**
- Build `s3stream` module: `./gradlew :s3stream:compileJava -x test` should pass.
- Run regular builds/integration tests to confirm no functional regressions.
- Optional: profile CPU/time of the monitor task with large object lists to confirm reduction.
        

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/compact/CompactionManager.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/compact/CompactionManager.java
@@ -55,7 +55,6 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Gezi-lzq:
Thank you for your contribution @JasirVoriya . It looks fine overall, but I noticed there are some CI errors that could be resolved.

JasirVoriya:
> Thank you for your contribution @JasirVoriya . It looks fine overall, but I noticed there are some CI errors that could be resolved.

Fixed it, thanks for the heads-up @Gezi-lzq .

woshigaopp (approved):
LGTM

PR #2987, opened 2025-11-04
perf(s3stream): fast path for block-start reads in StreamCache

JasirVoriya:
**Overview**
- Adds an O(1) start-index resolution when `startOffset` equals the block’s `startOffset` in `LogCache.StreamCache#searchStartIndex`.
- Keeps the cache-hit path first; falls back to binary search only when neither cache nor block-start match applies.
- No behavior change for sequential, cursor-based reads.

**Motivation**
- First reads from a cached block often start at the block’s beginning. Previously, these took the binary search path.

**Performance Impact**
- Cold-start/scan reads from block start:
  - Avoids `O(log N)` binary search and a small helper allocation.
  - Improves latency deterministically for large blocks.
- Cursor-based reads:
  - Unaffected; cache hits remain O(1).

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/compact/CompactionManager.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/compact/CompactionManager.java
@@ -55,7 +55,6 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


superhx:
@JasirVoriya Thank you for your PR. From the perspective of LogCache's usage scenarios, most calls to searchStartIndex will not hit startOffset == this.startOffset, so I think this optimization may not be very helpful.

JasirVoriya:
> @JasirVoriya Thank you for your PR. From the perspective of LogCache's usage scenarios, most calls to searchStartIndex will not hit startOffset == this.startOffset, so I think this optimization may not be very helpful.

@superhx Thanks for reviewing. That makes sense — given the typical LogCache access patterns, the gain from this path would indeed be limited.
I’ll close this PR for now.

PR #3015, opened 2025-11-15
fix(core): fix outbound bandwidth limiter token refill and metrics baseline

yx9o:
Outbound baseline bandwidth is configured as networkBaselineBandwidth * 5, but refillToken and metrics calculation were still using inbound baseline.
This caused dashboard outbound available bandwidth to always show negative values.
image

### DIFF
--- a/core/src/main/scala/kafka/log/stream/s3/DefaultS3Client.java
+++ b/core/src/main/scala/kafka/log/stream/s3/DefaultS3Client.java
@@ -140,11 +140,12 @@ public class DefaultS3Client implements Client {
             config.networkBaselineBandwidth() - (long) networkInboundRate.derive(
                 TimeUnit.NANOSECONDS.toSeconds(System.nanoTime()), NetworkStats.getInstance().networkInboundUsageTotal().get()));
         // Use a larger token pool for outbound traffic to avoid spikes caused by Upload WAL affecting tail-reading performance.
+        long outboundBaselineBandwidth = config.networkBaselineBandwidth() * 5;
         GlobalNetworkBandwidthLimiters.instance().setup(AsyncNetworkBandwidthLimiter.Type.OUTBOUND,
-            refillToken, config.refillPeriodMs(), config.networkBaselineBandwidth() * 5);
+            refillToken * 5, config.refillPeriodMs(), outboundBaselineBandwidth);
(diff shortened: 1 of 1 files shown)


superhx:
The negative values are caused by actual network exhaustion.

networkBaselineBandwidth * 5 is the max token size for the network bursting.

The metrics mean that the available bandwidth = network baseline - network consumption rate

yx9o:
In our actual production scenarios, outbound traffic is often larger than inbound traffic; for example, it's common for multiple groups to consume the same topic. Therefore, the baseline bandwidth of the outbound traffic should be adjusted to be larger than the baseline bandwidth of the inbound traffic. Otherwise, this metric will always be negative, making it difficult to determine the actual outbound bandwidth.

superhx:
> In our actual production scenarios, outbound traffic is often larger than inbound traffic; for example, it's common for multiple groups to consume the same topic. Therefore, the baseline bandwidth of the outbound traffic should be adjusted to be larger than the baseline bandwidth of the inbound traffic. Otherwise, this metric will always be negative, making it difficult to determine the actual outbound bandwidth.

In AWS / GCP / Azure, the inbound network and the outbound network are isolated. So setting the same baseline of the inbound network and outbound network is fine.

Is your environment's network shared by both inbound and outbound?

yx9o:
Um, we are using it in a self-built data center, where the bandwidth is relatively sufficient. The outbound traffic is many times that of the inbound traffic. We would like to set different baseline bandwidths for the inbound and outbound traffic. Do you have any good suggestions for modification?

superhx:
> Um, we are using it in a self-built data center, where the bandwidth is relatively sufficient. The outbound traffic is many times that of the inbound traffic. We would like to set different baseline bandwidths for the inbound and outbound traffic. Do you have any good suggestions for modification?

If the inbound and outbound traffic in a self-built data center share the same bandwidth, perhaps we should create an AsyncNetworkBandwidthLimiter that is shared by both inbound and outbound traffic.

yx9o:
My idea is to keep the peak value of out as baseline * 5, and to differentiate the baseline of out from that of in. For example, out could be baseline * 2 (this 2 is a coefficient, the default is 1, and it would be best if it were configurable).

PR #3046, opened 2025-11-25
fix(core): improve resource teardown in S3Client shutdown sequence

yx9o:
Hi community, we encountered the following problem during verification:
The `DefaultS3Client#shutdown()` method previously missed shutting down the `S3StreamThreadPoolMonitor`, which resulted in residual logging after the client shutdown was declared complete.


superhx:
S3StreamThreadPoolMonitor.MONITOR_SCHEDULED is a daemon thread. It is expected to be auto shutdown when the process exits.

yx9o:
image
The problem we are encountering now is that the shutdown method has been executed and there is a shutdown successfully log output, but the ThreadPoolMonitor log is still being output continuously.

superhx:
@yx9o It is as expected; continuous printing indicates that the node has not been turned off.

PR #3075, opened 2025-12-01
feat(core): Replace fixed parameters with environment variables

yx9o:
Replace `kafka.log.stream.s3.network.ControllerRequestSender#MAX_RETRY_DELAY_MS` with a value that can be set via environment variables.

### DIFF
--- a/core/src/main/scala/kafka/log/stream/s3/network/ControllerRequestSender.java
+++ b/core/src/main/scala/kafka/log/stream/s3/network/ControllerRequestSender.java
@@ -52,7 +52,7 @@ public class ControllerRequestSender {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ControllerRequestSender.class);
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/log/stream/s3/network/ControllerRequestSender.java
```diff
@@ -51,7 +52,7 @@ public class ControllerRequestSender {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ControllerRequestSender.class);
 
-    private static final long MAX_RETRY_DELAY_MS = 10 * 1000; // 10s
+    private static final long MAX_RETRY_DELAY_MS = Systems.getEnvLong("AUTOMQ_MAX_RETRY_DELAY_MS", 10L * 1000); // 10s
```
superhx: Maybe the env name could be more specific such as AUTOMQ_CONTROLLER_REQUEST_MAX_RETRY_DELAY_MS

yx9o: It has been adjusted. Please review it again. Thank you.

superhx (approved):

PR #3107, opened 2025-12-15
docs: Improve contributing guide for clearer onboarding

lobsangshakya:
@KaimingWan, I kept the original content temporarily to make review and comparison easier. Once the changes are approved, I will remove the reference to keep the guide clean.


KaimingWan (changes requested):
Thank you for your contribution. The new version is indeed more concise and readable. However, I noticed that you removed the localstack section, which could make it more challenging for beginners to set up and run AutoMQ locally. Given that AutoMQ depends on an object storage service to function properly, what are your thoughts on this area?

PR #3175, opened 2026-01-22
feat(automq-shell): support custom port in topo.yaml for Node

KaimingWan:
support custom port in topo.yaml for Node


automq-shell/src/test/java/com/automq/shell/model/NodePortTest.java
```diff
@@ -0,0 +1,54 @@
+/*
```
superhx, line 1: If you want to deploy multiple controller / broker in the same machine, you should consider the conflict of all ports

PR #3208, opened 2026-02-06
fix(log): improve error handling and optimize active segment manageme…

1sonofqiu:
…nt in append

comment to ping reviewers.*


core/src/main/scala/kafka/log/streamaspect/CachedLogSegments.java
superhx, line 39: We should reset the activeSegment when the segments are changed in all condition.

PR #3246, opened 2026-03-09
feat(extension): add SPI-based enterprise API routing

Gezi-lzq:


core/src/main/scala/kafka/server/streamaspect/ElasticKafkaApis.scala
```diff
@@ -91,6 +92,35 @@ class ElasticKafkaApis(
 
   private var trafficInterceptor: TrafficInterceptor = new NoopTrafficInterceptor(this, metadataCache)
   private var snapshotAwaitReadySupplier: Supplier[CompletableFuture[Void]] = () => CompletableFuture.completedFuture(null)
+  private val brokerExtensionContext: BrokerExtensionContext = new BrokerExtensionContext {
```
superhx: Splitting it into a separate Inner Class makes the declaration of member variables more concise.

Issue #3111, opened 2025-12-16
[BUG] Kafka broker fails to start when enabling OTLP HTTP metrics exporter


kincoy:
### AutoMQ Version

1.6.0-strimzi

### Operating System

kubernetes 1.24

### Installation Method

other

### Hardware Configuration

16 core, 32 GB memory

### Other Relevant Software

kafka 3.9.0

### What Went Wrong?

Enabling the OTLP metrics exporter causes the broker to fail during startup. The process logs a warning about multiple HttpSenderProvider implementations and then crashes with:

java.lang.AbstractMethodError in io.opentelemetry.exporter.sender.okhttp.internal.OkHttpHttpSenderProvider while building OtlpHttpMetricExporter (via HttpExporterBuilder.build). Broker startup aborts and shutdown subsequently hits NPEs due to incomplete initialization.

### What Should Have Happened Instead?

The broker should start normally and periodically export metrics via OTLP/HTTP to the configured endpoint without crashing.

### Steps to Reproduce

1. Configure metrics exporter: `s3.telemetry.metrics.exporter.uri`
2. Start the broker (Kubernetes/**Strimzi** environment).
3. Observe the broker exits during startup with the error below.

### Additional Information

relevant logs:

2025-12-16 10:01:43,808 INFO s3.block.cache.size is not set, using 536870912 as the default value (kafka.server.KafkaConfig) [main]
2025-12-16 10:01:43,809 INFO s3.wal.upload.threshold is not set, using 357913941 as the default value (kafka.server.KafkaConfig) [main]
All of the log directories are already formatted.
KRaft storage formatting is done

Preparing Kafka Agent configuration

+ exec /usr/bin/tini -w -e 143 -- /opt/kafka/bin/kafka-server-start.sh /tmp/strimzi.properties
2025-12-16 10:01:46,291 INFO Starting KafkaAgent with sslKeyStorePath=/tmp/kafka/cluster.keystore.p12 and sslTrustStore=/tmp/kafka/cluster.truststore.p12 (io.strimzi.kafka.agent.KafkaAgent) [main]
2025-12-16 10:01:46,310 INFO Logging initialized @840ms to org.eclipse.jetty.util.log.Slf4jLog (org.eclipse.jetty.util.log) [main]
[...]

PR #3252, opened 2026-03-20
fix: Prevent broker startup crash when using GRPC for metrics to prevent broker crash

### LINKED ISSUE #3217: [BUG] OTLP gRPC metrics exporter crashes on startup — GrpcSenderProvider missing from classpath
When configuring the OTLP metrics exporter with `protocol=grpc`, the broker crashes on startup with:

```
java.lang.IllegalStateException: No GrpcSenderProvider found on classpath.
Please add dependency on opentelemetry-exporter-sender-okhttp or opentelemetry-exporter-sender-grpc-upstream
```

PR #3124 added `opentelemetry-exporter-sender-jdk` and excluded `opentelemetry-exporter-sender-okhttp` from `automq-metrics`. `sender-jdk` only provides `HttpSenderProvider` via SPI — it does not provide `GrpcSenderProvider`. The only module that registers `GrpcSenderProvider` is `sender-okhttp`, which is explicitly excluded.

The `OTLPProtocol` enum still declares `GRPC` and `OTLPMetricsExporter` has a `case GRPC` branch that compiles (the API classes exist in `exporter-otlp`), but fails at runtime because no SPI implementation is available.

manmohak07:
fixes #3217 

The changes fix a bug where configuring the OTLP metrics exporter to use the grpc protocol would cause the broker to immediately crash during startup.

A recent update (PR #3124 ) removed the OkHttp network library to rely on the built-in JDK client instead. This worked fine for sending metrics over standard HTTP, but it removed the ability to send metrics over GRPC.

As a result, if a user configured their metrics exporter to use GRPC, the system lacked the tool needed to send the data and crashed on startup with an `IllegalStateException: No GrpcSenderProvider found` error. 

The changes re-introduce the `opentelemetry-exporter-sender-okhttp` dependency to the `automq-metrics` module. This naturally restores the missing GRPC tool for OpenTelemetry, allowing the broker to start safely and transmit metrics over GRPC again.

It should now successfully boot up and begin exporting telemetry without throwing a missing classpath exception.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2352,7 +2352,6 @@ project(':automq-metrics') {
     api libs.opentelemetryExporterLogging
     api libs.opentelemetryExporterProm
     api libs.opentelemetryExporterOTLP
-    api libs.opentelemetryExporterSenderJdk
     api libs.opentelemetryExporterSenderOkhttp
     api libs.opentelemetryJmx
 
--- a/build.gradle
+++ b/build.gradle
@@ -2341,11 +2341,7 @@ project(':automq-metrics') {
     configProperties = checkstyleConfigProperties("import-control-server.xml")
   }
 
-  configurations {
-    all {
-      exclude group: 'io.opentelemetry', module: 'opentelemetry-exporter-sender-okhttp'
-    }
-  }
+
     
   dependencies {
     // OpenTelemetry core dependencies
@@ -2357,6 +2353,7 @@ project(':automq-metrics') {
     api libs.opentelemetryExporterProm
     api libs.opentelemetryExporterOTLP
     api libs.opentelemetryExporterSenderJdk
+    api libs.opentelemetryExporterSenderOkhttp
     api libs.opentelemetryJmx
 
     // Logging dependencies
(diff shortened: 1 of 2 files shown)


Gezi-lzq (commented):
Thanks for the fix! The change correctly restores gRPC support by re-introducing `sender-okhttp` with the proper version alignment.

One suggestion: now that `sender-okhttp` is back (which provides both `HttpSenderProvider` and `GrpcSenderProvider`), `sender-jdk` becomes redundant. Keeping both means two `HttpSenderProvider` implementations on the classpath, which causes OTel to emit a WARNING log at startup:

```
WARN Multiple HttpSenderProvider found. Please include only one, or specify preference setting ...
```

Consider removing `opentelemetryExporterSenderJdk` from the dependencies to keep things clean. `sender-okhttp` alone covers both `http` and `grpc` protocols.

Gezi-lzq:
One more thing — while investigating this PR, we traced the full history (#3111 → #3124 → #3217 → this PR) and found that the original `AbstractMethodError` in #3111 was actually caused by an OTel version mismatch in the Strimzi image build, not by AutoMQ's dependency declarations.

The Strimzi `kafka-thirdparty-libs/pom.xml` bundles OTel 1.34.1, while AutoMQ ships 1.40.0. Both end up in `/opt/kafka/libs/`, and the mixed versions cause `AbstractMethodError` at runtime.

I've opened #3260 with the root cause analysis and two proposed fixes on the Strimzi side. Would you be interested in also addressing the Strimzi `pom.xml` as part of this effort? The fix there would be either upgrading OTel to 1.40.0 or removing the duplicate OTel deps from thirdparty-libs entirely.

manmohak07:
Sure, I think we can do it.

Gezi-lzq (dismissed):

Gezi-lzq:
Thanks for sharing this; based on my current understanding, the dependency issue discussed here has largely been resolved, so I’ll close this PR for now.
Thanks again for your patient contribution.

Issue #3253, opened 2026-03-20
[BUG] Brokers/Controllers enter crashloop when OTLP HTTP metrics exporter is enabled


gjpin:
### AutoMQ Version

1.6.4-strimzi

### Operating System

Bottlerocket 1.56

### Installation Method

other

### Hardware Configuration

- 4 cores
- 16GB of memory

### Other Relevant Software

- Kafka 3.9.0
- Strimzi 0.47
- Kubernetes 1.35

### What Went Wrong?

AutoMQ crashes with the following error in the logs:
```
kafka java.lang.AbstractMethodError: Receiver class io.opentelemetry.exporter.sender.okhttp.internal.OkHttpHttpSenderProvider does not define or inherit an implementation of the resolved method 'abstract io.opentelemetry.exporter.internal.http.HttpSender createSender(java.lang.String, io.opentelemetry.exporter.internal.compression.Compressor, boolean, java.lang.String, long, long, java.util.function.Supplier, io.opentelemetry.exporter.internal.auth.Authenticator, io.opentelemetry.sdk.common.export.RetryPolicy, javax.net.ssl.SSLContext, javax.net.ssl.X509TrustManager)' of interface io.opentelemetry.exporter.internal.http.HttpSenderProvider.
```

### What Should Have Happened Instead?

AutoMQ should not crash and should send the metrics in OTLP format to the defined OTLP endpoint.

### Steps to Reproduce

Configure AutoMQ to export metrics via OTLP with the following configuration:
```
s3.telemetry.metrics.exporter.type=otlp
s3.telemetry.exporter.otlp.protocol=http
s3.telemetry.exporter.otlp.endpoint=http://${endpoint}:9090/api/v1/otlp/v1/metrics
```

Source: https://github.com/automq/automq/wiki/Integrating-Metrics-with-Prometheus

### Additional Information

Related issue: https://github.com/AutoMQ/automq/issues/3111
Relevant PR that should fix this issue: https://github.com/AutoMQ/automq/pull/3124
Image in use: docker.io/automqinc/automq:1.6.4-strimzi

manmohak07:
`sender-okhttp` is not in the global` force()` list. So if any transitive dependency brings in an older version, it can win the resolution.

According to this issue (https://github.com/gradle/gradle/issues/7317) transitive resolution is not guaranteed.

I think we can add these two lines.

```java
libs.opentelemetryExporterSenderOkhttp,
libs.opentelemetryExporterSenderJdk,
```

[...]

PR #3268, opened 2026-03-30
fix: replace sender-okhttp with sender-grpc-managed-channel for Strimzi compatibility

### LINKED ISSUE #3260: [BUG] OpenTelemetry version mismatch between Strimzi thirdparty-libs and AutoMQ causes startup crash
## Summary

The Strimzi image bundles OpenTelemetry 1.34.1 via `kafka-thirdparty-libs`, while the AutoMQ Kafka tarball ships OpenTelemetry 1.40.0. Both sets of jars end up in the flat `/opt/kafka/libs/` directory, causing version conflicts at runtime.

## Related Issues

- #3111 — Original bug: `AbstractMethodError` in `OkHttpHttpSenderProvider` when OTLP HTTP exporter is enabled
- #3124 — Workaround: excluded `sender-okhttp`, added `sender-jdk` in AutoMQ
- #3217 — Regression: gRPC protocol crashes because `GrpcSenderProvider` was removed along with `sender-okhttp`
- #3252 — Fix: re-introduces `sender-okhttp` in AutoMQ (correct for AutoMQ standalone, but doesn't address the Strimzi conflict)

## Root Cause

The Strimzi image build (`AutoMQ/strimzi-kafka-operator`) copies all jars into `/opt/kafka/libs/`:

| Source | OTel Version | Key Jars |
|--------|-------------|----------|
| AutoMQ tarball | **1.40.0** | exporter-otlp, sdk, sender-okhttp, sender-jdk |
| Strimzi thirdparty-libs (`pom.xml`) | **1.34.1** | exporter-otlp, sdk, sender-jdk (1.34.1-alpha), sender-grpc-managed-channel |

When two versions of the same OTel artifact coexist on the classpath, the JVM may load a class from one version and an interface from another. Specifically:

1. `HttpExporterBuilder` (1.40.0) calls `HttpSenderProvider.createSender(...)` with the 1.40.0 method signature
2. `ServiceLoader` may resolve `OkHttpHttpSenderProvider` from the 1.34.1 jar (or vice versa)
3. The method signature changed between 1.34.1 and 1.40.0 → `AbstractMethodError`

The `find-classpath-collision.sh` script in Strimzi catches class-level collisions but not same-artifact version mismatches, so this slips through the build.

## Why previous fixes were incomplete

- **#3124** excluded `sender-okhttp` and added `sender-jdk` in AutoMQ to avoid the `AbstractMethodError`. This worked for `protocol=http` but broke `protocol=grpc` because `GrpcSenderProvider` is only provided by `sender-okhttp`.
- **#3252** re-introduces `sender-okhttp` in AutoMQ, which is correct for standalone AutoMQ. But in the Strimzi environment, the version mismatch between AutoMQ's 1.40.0 jars and Strimzi's 1.34.1 jars remains the root cause.

The real fix needs to happen in `AutoMQ/strimzi-kafka-operator`'s `kafka-thirdparty-libs/pom.xml`.

### Option A: Upgrade OTel version to match AutoMQ (minimal change)

Align `kafka-thirdparty-libs/3.9.x/pom.xml` and `4.0.x/pom.xml` to 1.40.0:

```xml
<opentelemetry.version>1.40.0</opentelemetry.version>
<opentelemetry-alpha.version>1.40.0-alpha</opentelemetry-alpha.version>
```

- Pros: minimal diff, both sides use the same version, no behavioral change.
- Cons: still ships duplicate jars (one from AutoMQ tarball, one from thirdparty-libs). Needs to be re-aligned whenever AutoMQ upgrades OTel.

### Option B: Remove duplicate OTel dependencies from thirdparty-libs (thorough)

Since the AutoMQ tarball already ships a complete set of OTel jars (sdk, exporters, senders), the thirdparty-libs `pom.xml` should stop declaring them. Only keep what AutoMQ does not provide (e.g., `opentelemetry-kafka-clients-2.6` for Strimzi tracing, `opentelemetry-sdk-extension-autoconfigure`).

```xml
<!-- Remove these from pom.xml (already in AutoMQ tarball): -->
<!-- opentelemetry-exporter-otlp -->
<!-- opentelemetry-exporter-sender-jdk -->
<!-- opentelemetry-exporter-sender-grpc-managed-channel -->
<!-- grpc-netty-shaded -->

<!-- Keep these (Strimzi-specific, not in AutoMQ tarball): -->
<!-- opentelemetry-sdk-extension-autoconfigure -->
<!-- opentelemetry-kafka-clients-2.6 -->
```

- Pros: eliminates duplication at the source, no version drift risk.
- Cons: larger change, requires verifying exactly which OTel jars the AutoMQ tarball provides. Strimzi's tracing agent dependencies (`autoconfigure`, `kafka-clients-2.6`) may also need version alignment.

## Environment

- AutoMQ OTel SDK version: 1.40.0 (`gradle/dependencies.gradle`)
- Strimzi thirdparty-libs OTel version: 1.34.1 (`kafka-thirdparty-libs/3.9.x/pom.xml` and `4.0.x/pom.xml`)

rafaelsoutto:
Resolves AutoMQ/automq#3260 and AutoMQ/automq#3253.

The Strimzi thirdparty-libs bundle includes OpenTelemetry 1.34.1 with sender-okhttp, which conflicts with AutoMQ's OTel 1.40.0. This caused AbstractMethodError on OkHttpHttpSenderProvider.createSender() at runtime.

Changes:
- Keep sender-okhttp excluded from automq-metrics dependencies
- Add sender-grpc-managed-channel + grpc-netty-shaded for gRPC transport
- Keep sender-jdk for HTTP/protobuf transport
- Pin sender versions in the global force() list to prevent version drift
- Add build-strimzi-image.sh convenience script

comment to ping reviewers.*


Gezi-lzq (approved):
LGTM

PR #3312, opened 2026-04-14
feat: add cluster events framework with protobuf-based event publishing

superhx:
Introduce a cluster-level event system that publishes structured events to the `__automq_cluster_events` internal topic using CloudEvents + protobuf.

**Core infrastructure:**
- `ClusterEventPublisher`: async, best-effort producer that wraps events as CloudEvents and writes them to the internal topic
- `ClusterEventsReader`: consumer-based reader for retrieving events
- `ClusterEventsCommand`: CLI tool (`kafka-cluster-events.sh`) for reading and displaying cluster events

**RequestErrorEvent pipeline:**
- Detects sustained request error patterns (e.g. auth failures, authorization errors) on the broker and publishes them as events, enabling the control plane to surface risky or misconfigured clients
- Integrated into `RequestChannel` and `KafkaApis` to capture errors on the broker request path
- `ResourceErrorExtractor` parses per-resource error codes from Kafka API responses (Produce, Fetch, Metadata, etc.)
- `RequestErrorAccumulator` aggregates errors in-memory by (apiKey, errorCode, resource) with client IP/ID tracking, then periodically flushes accumulated buckets as `RequestErrorEvent` records

### DIFF
--- a/core/src/main/java/kafka/server/RequestErrorAccumulator.java
+++ b/core/src/main/java/kafka/server/RequestErrorAccumulator.java
@@ -1,0 +1,135 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kafka.server;
+
+import org.apache.kafka.clients.admin.ClusterEventPublisher;
+import org.apache.kafka.clients.admin.RequestErrorEventData;
+import org.apache.kafka.common.protocol.ApiKeys;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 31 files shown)


woshigaopp (commented):
(deleted - moved to inline comment)

core/src/main/java/kafka/server/RequestErrorAccumulator.java
```diff
@@ -0,0 +1,135 @@
...
+    public void record(short apiKey, short errorCode, String resource,
+                       String clientIp, String clientId) {
+        ErrorKey key = new ErrorKey(apiKey, errorCode, resource);
+        ErrorBucket bucket = buckets.computeIfAbsent(key, k -> new ErrorBucket());
+        bucket.count.increment();
+        if (bucket.clientIps.size() < MAX_CLIENTS) {
+            bucket.clientIps.add(clientIp);
+        }
+        if (bucket.clientIds.size() < MAX_CLIENTS) {
+            bucket.clientIds.add(clientId);
```
woshigaopp, line 80: This records every non-zero error code, but many are part of normal Kafka protocol flow and not "risky" — e.g. `REBALANCE_IN_PROGRESS`, `NOT_COORDINATOR`, `UNKNOWN_TOPIC_OR_PARTITION` (first metadata fetch for a new topic), `NOT_LEADER_OR_FOLLOWER` (during leader election), etc.

On a busy cluster this will generate a large volume of `RequestErrorEvent` noise, making it harder for the control plane to surface genuinely risky patterns (auth/authz failures) — which seems to be the stated goal of this pipeline.

A few options:

1. **Allowlist**: Only record error codes that represent actual risk signals (`TOPIC_AUTHORIZATION_FAILED`, `GROUP_AUTHORIZATION_FAILED`, `CLUSTER_AUTHORIZATION_FAILED`, `SASL_AUTHENTICATION_FAILED`, etc.).
2. **Denylist**: Exclude well-known benign/transient errors (`NOT_LEADER_OR_FOLLOWER`, `REBALANCE_IN_PROGRESS`, `NOT_COORDINATOR`, `UNKNOWN_TOPIC_OR_PARTITION`, `COORDINATOR_NOT_AVAILABLE`, etc.).
3. **Keep as-is but make it configurable**: If the design is intentionally broad (let the control plane decide), document that expectation and add a configurable filter so operators can tune the noise level.

What was the intent here?

superhx: All ErrorCodes need to be recorded.

woshigaopp (approved):

PR #3316, opened 2026-04-14
fix(s3stream): fix double-release race in LogCache.tryMerge (#3307)

superhx:
## Bug

A rare `IllegalReferenceCountException: refCnt: 0, decrement: 1` occurs when `LogCacheBlock.free()` tries to release a `StreamRecordBatch` whose refCnt is already 0.

Root cause: `tryMerge` performs the expensive `mergeBlock` call outside the write lock for performance. During that window, `clearStreamRecords` can acquire the write lock and call `block.free(streamId)`, which removes a `StreamCache` from the block and schedules an async release of all its `StreamRecordBatch` records (refCnt → 0). When `tryMerge` then re-acquires the write lock, it only checks that `left`/`right` are still the same block objects in `blocks` — it does not detect that the blocks were mutated. So the merged `newBlock` (which shares the same record references via `mergeBlock`) replaces `left`/`right` in `blocks`. Later, when `newBlock.free()` is called, it tries to release records that are already at refCnt 0, causing the crash.

## Fix

Add a `freeOpsModCount` field to `LogCacheBlock` that is incremented whenever the block is mutated by a free operation (`free()` or `free(long streamId)`). In `tryMerge`, snapshot both blocks' `freeOpsModCount` inside the same write lock that selects `left` and `right`. After `mergeBlock` completes, re-acquire the write lock and verify the counts are unchanged before committing the swap. If either block was mutated in between, the merge is aborted and `newBlock` is simply discarded — preventing stale record references from entering `blocks` and being double-freed.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/cache/LogCache.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/cache/LogCache.java
@@ -312,6 +312,8 @@ public class LogCache {
         for (; ; ) {
             LogCacheBlock left;
             LogCacheBlock right;
+            long leftFreeOpsModCount;
+            long rightFreeOpsModCount;
             writeLock.lock();
             try {
                 if (blocks.size() <= MERGE_BLOCK_THRESHOLD || mergeStartIndex + 1 >= blocks.size()) {
@@ -326,6 +328,8 @@ public class LogCache {
                     mergeStartIndex++;
                     continue;
                 }
+                leftFreeOpsModCount = left.freeOpsModCount;
+                rightFreeOpsModCount = right.freeOpsModCount;
             } finally {
                 writeLock.unlock();
             }
@@ -343,6 +347,8 @@ public class LogCache {
                 if (blocks.size() > mergeStartIndex + 1
                     && blocks.get(mergeStartIndex) == left
                     && blocks.get(mergeStartIndex + 1) == right
+                    && left.freeOpsModCount == leftFreeOpsModCount
+                    && right.freeOpsModCount == rightFreeOpsModCount
                 ) {
                     blocks.set(mergeStartIndex, newBlock);
                     blocks.remove(mergeStartIndex + 1);
@@ -428,6 +434,7 @@ public class LogCache {
         private final AtomicLong size = new AtomicLong();
         private final List<FreeListener> freeListeners = new ArrayList<>();
         volatile boolean free;
+        long freeOpsModCount;
         private RecordOffset lastRecordOffset;
 
         public LogCacheBlock(long maxSize, int maxStreamCount) {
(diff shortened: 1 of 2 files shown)


Gezi-lzq (approved):

superhx:
A a cross-cache ByteBuf lifetime corruption bug.

  SnapshotReadCache does not directly generate the corrupted SSO. The SSO is generated from DeltaWALCache. The connection is Netty pooled ByteBuf wrapper reuse: a
  stale double-release in SnapshotReadCache can accidentally release a live DeltaWALCache record after both records reuse the same recycled ByteBuf wrapper.

  Workflow

  1. SnapshotReadCache owns a LogCache used for snapshot/read caching.
  2. Two free snapshot cache blocks are selected by LogCache.tryMerge().
  3. tryMerge() releases the write lock while it performs expensive merge work.
  4. During that unlocked window, SnapshotReadCache calls cache.clearStreamRecords(streamId).
  5. clearStreamRecords() removes that stream from the old snapshot blocks and asynchronously releases its StreamRecordBatch records.
  6. One old snapshot record, Record-A, releases its encoded ByteBuf wrapper to refCnt=0.
  7. Netty recycles that released wrapper.
  8. A later append record in DeltaWALCache, Record-B, reuses the same wrapper.
  9. The old tryMerge() code fails to notice that the source snapshot blocks were mutated during the unlocked merge window.
  10. It swaps a merged snapshot block back into the cache with stale references to already-released Record-A.
  11. Later, the merged snapshot block is freed.
  12. It releases stale Record-A again.
  13. Because Record-A.encoded and live Record-B.encoded now point to the same recycled wrapper object, this second release false-releases Record-B.
  14. Netty recycles the wrapper again.
  15. Another later record, Record-C from a foreign stream, reuses that wrapper.
  16. DeltaWALCache still logically contains Record-B in the correct stream list and offset position, but byte-backed methods now read Record-C bytes.
  17. DefaultUploadWriteAheadLogTask builds an SSO from DeltaWALCache.
  18. ObjectWriter.DataBlock trusts the stream grouping and does not validate that every encoded record header still matches the stream id.
  19. The object index says “this block belongs to Stream-B”, but the physical data contains one foreign Record-C.
  20. On read, the stream filter skips the foreign record. The next real record for the indexed stream has a higher offset, so the reader reports:

  Continuous check failed, expected offset: X, actual: Y

  Fix Relation

  Commit 2e69641db35f3b320aba0725d9871306f5d4d65b fixes the trigger by adding a mutation counter to LogCacheBlock. tryMerge() now aborts the swap if
  clearStreamRecords() or free() mutates either source block during the unlocked merge window. This prevents stale already-released snapshot records from re-entering
  the cache and being released again later.

PR #3322, opened 2026-04-15
fix: force jaxb-api to 2.3.1 to avoid duplicate versions in release packaging

Gezi-lzq:
## Problem

Release tarball `libs/` contains both `jaxb-api-2.2.12.jar` and `jaxb-api-2.3.1.jar` because:
- `:core` resolves `jaxb-api` to `2.2.12` (from `jackson-module-jaxb-annotations:2.20.0` transitive dep)
- `:trogdor`/`:connect:runtime`/`:connect:mirror` resolve to `2.3.1` (direct `implementation libs.jaxbApi`)

Each module resolves independently, and `releaseTarGz` copies all runtimeClasspaths into a flat `libs/` directory. `duplicatesStrategy EXCLUDE` only deduplicates same-name files, so both versions coexist.

## Fix

Add `libs.jaxbApi` to the global `configurations.all.resolutionStrategy.force` block, ensuring all modules resolve `jaxb-api` to the declared `2.3.1` version.

## Verification

```
# Before: :core resolves jaxb-api to 2.2.12
# After:
javax.xml.bind:jaxb-api:2.2.12 -> 2.3.1 (forced)
javax.xml.bind:jaxb-api:2.2.11 -> 2.3.1 (forced)
javax.xml.bind:jaxb-api:2.2.2  -> 2.3.1 (forced)

# dependant-libs: only jaxb-api-2.3.1.jar ✅
```

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -114,6 +114,7 @@ versions += [
   scalaLogging: "3.9.5",
   jaxAnnotation: "1.3.2",
   jaxb: "2.3.1",
+  jaxbRI: "2.3.1", // JAXB Reference Implementation version matching jaxb-api 2.3.1
   jaxrs: "2.1.1",
   jfreechart: "1.0.0",
   jopt: "5.0.4",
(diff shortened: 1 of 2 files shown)


gradle/dependencies.gradle
```diff
@@ -114,6 +114,7 @@ versions += [
   scalaLogging: "3.9.5",
   jaxAnnotation: "1.3.2",
   jaxb: "2.3.1",
+  jaxbRI: "2.3.0.1", // JAXB Reference Implementation version matching jaxb-api 2.3.1
```
superhx: 2.3.1 or 2.3.0.1?

superhx (dismissed):

superhx (approved):

PR #3323, opened 2026-04-15
fix: buffer over-allocation in AutoBalancerMetrics.bodySize()

cnkLmxf:
### Problem
  `AutoBalancerMetrics.bodySize()` incorrectly uses bit-width constants (`Integer.SIZE`, `Byte.SIZE`, `Double.SIZE`) to calculate buffer size in bytes. Since these
  constants represent bits (32, 8, 64) rather than bytes (4, 1, 8), the method returns 8x the actual required size.

  Example: With 10 metrics, `bodySize()` returns 1000 bytes:
  - `(32 + (8 + 64) * 10) = 1000` bits treated as bytes

  Instead of the correct 94 bytes:
  - `(4 + (1 + 8) * 10) = 94` bytes

  ### Solution
  Replace bit-width constants with byte-width constants:
  - `Integer.SIZE` → `Integer.BYTES` (4 bytes)
  - `Byte.SIZE` → `Byte.BYTES` (1 byte)
  - `Double.SIZE` → `Double.BYTES` (8 bytes)

  ### Impact
  This reduces memory allocation when `encode()` creates the ByteBuffer for serializing metrics to the `__auto_balancer_metrics` internal topic.

  ### Testing
  - Added AutoBalancerMetricsTest.testBodySizeCalculation() verifying:
    - Empty/single/multiple metrics size calculation
    - bodySize matches actual bytes written by writeBody()
    - 10 metrics: 94 bytes (not 1000)
  - All existing tests pass: MetricSerdeTest, MetricsUtilsTest

### DIFF
--- a/core/src/main/java/kafka/autobalancer/metricsreporter/metric/AutoBalancerMetrics.java
+++ b/core/src/main/java/kafka/autobalancer/metricsreporter/metric/AutoBalancerMetrics.java
@@ -98,7 +98,7 @@ public abstract class AutoBalancerMetrics {
     }
 
     public int bodySize() {
-        return Integer.SIZE + (Byte.SIZE + Double.SIZE) * metricValueMap.size();
+        return Integer.BYTES + (Byte.BYTES + Double.BYTES) * metricValueMap.size();
     }
 
     public ByteBuffer writeBody(ByteBuffer buffer) {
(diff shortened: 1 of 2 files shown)


Gezi-lzq (approved):
LGTM

PR #3325, opened 2026-04-17
refactor: encapsulate KVImage/KVDelta map access to enforce KVKey type safety

superhx:
## Motivation

After the KV key type changed from `String` to `KVKey`, callers that still pass a raw `String` to `Map<KVKey, ByteBuffer>.get(Object)` compile without error but silently return `null` at runtime (since `String.equals(KVKey)` is always `false`).

This was already causing a bug in `FailoverListener`, where failover context changes were never detected. This bug is caused by https://github.com/AutoMQ/automq/pull/3306

## Changes

- **KVImage**: make `kvs()` package-private; add `public forEach(BiConsumer<KVKey, ByteBuffer>)` for safe iteration
- **KVDelta**: make `changedKV()` package-private; add `public getChangedKV(KVKey)` for type-safe lookup
- **FailoverListener**: fix bug — was passing `String` to `Map<KVKey,...>.get()`, now uses `getChangedKV(KVKey.of(...))`
- **DefaultRouterChannelProvider**: migrate to `getChangedKV()`
- **KVImageNode**: migrate to `forEach()`

By hiding the raw `Map`, callers are forced to go through typed APIs, preventing this class of bug at compile time.

### DIFF
--- a/core/src/main/java/kafka/automq/failover/FailoverListener.java
+++ b/core/src/main/java/kafka/automq/failover/FailoverListener.java
@@ -21,7 +21,7 @@ package kafka.automq.failover;
 
 import kafka.automq.utils.JsonUtils;
 
-import org.apache.kafka.image.KVDelta;
+import org.apache.kafka.controller.stream.KVKey;
 import org.apache.kafka.image.MetadataDelta;
 import org.apache.kafka.image.MetadataImage;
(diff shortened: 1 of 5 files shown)


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3324

allenzhu (approved):

PR #3331, opened 2026-04-22
feat(routerchannel): support api for querying RouterChannelEpoch

allenzhu:


automq-shell/src/main/java/com/automq/shell/commands/cluster/Describe.java
```diff
@@ -66,6 +66,8 @@ public Integer call() throws Exception {
             for (NodeMetadata nodeMetadata : nodeMetadataList) {
                 System.out.println(nodeMetadata);
             }
+            GetNodesResult.RouterChannelEpoch epoch = rst.routerChannelEpoch().get();
+            System.out.println("RouterChannelEpoch: " + epoch);
```
superhx: Test code?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4976,6 +4965,7 @@ void handleFailure(Throwable throwable) {
     @Override
     public GetNodesResult getNodes(Collection<Integer> nodeIdList, GetNodesOptions options) {
```
superhx, line 4966: Maybe we could change the return to CompletableFuture<GetNodesResult>

automq-shell/src/main/java/com/automq/shell/commands/cluster/Describe.java
allenzhu: To make the Describe command output epoch results in the command-line terminal

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
allenzhu: In order to maintain backward compatibility, so we use the getNodes API in some places to prevent compatibility issues, and therefore keep its return type unchanged.

clients/src/main/resources/common/message/AutomqGetNodesResponse.json
```diff
@@ -78,6 +78,20 @@
           "about": "The tag value."
         }
       ]}
-    ]}
+    ]},
+    {
+      "name": "RouterChannelEpoch",
+      "type": "RouterChannelEpoch",
+      "versions": "0+",
+      "taggedVersions": "0+",
```
superhx, line 86: Refer to other request / response evolution. And note the AutomqGetNodes Request/Response validVersions / flexibleVersions also need to be changed

superhx (approved):

PR #3335, opened 2026-04-23
fix: exclude transitive kafka-clients from cloudevents-kafka (#3334)

superhx:
## Problem
The `cloudevents-kafka` dependency transitively pulls in `kafka-clients:3.0.0`, which conflicts with the project's own `kafka-clients:3.9.1`. This results in two kafka-clients jars in the release tarball.

## Solution
Exclude `org.apache.kafka:kafka-clients` from `cloudevents-kafka` in both `:clients` and `:tools` projects.

## Verification
After the fix, `./gradlew releaseTarGz` produces a tarball containing only `kafka-clients-3.9.1.jar`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1721,7 +1721,9 @@ project(':clients') {
     implementation libs.protobuf
 
     // AutoMQ inject start
-    implementation libs.cloudeventsKafka
+    implementation (libs.cloudeventsKafka) {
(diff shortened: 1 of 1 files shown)


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3334

allenzhu (approved):

PR #3339, opened 2026-04-27
fix: prevent data loss caused by DataBlockIndex.endOffsetDelta int erflow during compaction

superhx:
Bug Description:
DataBlockIndex.endOffsetDelta is stored as a 4-byte signed int (max 2^31-1). When a LogCacheBlock accumulates records for a single stream where endOffset - startOffset exceeds Integer.MAX_VALUE, the WAL upload writes a DataBlockIndex with an overflowed (negative) endOffsetDelta. This causes DataBlockIndex.endOffset() to return a value less than startOffset.

During compaction, filterInvalidStreamDataBlocks checks block.getEndOffset() <= stream.startOffset(). The corrupted endOffset satisfies this condition, so the block is removed. The object still has blocks for other streams, so it gets included in the compaction plan and deleted from S3. The affected stream's data is permanently lost.

This primarily affects compact topics (e.g. __consumer_offsets) where log cleaning produces .cleaned segments with sparse offset ranges. The DeltaWALUploadTask batches multiple such segments into a single object, creating a combined offset delta that crosses the int overflow threshold.

How it is fixed:
1. LogCache overflow prevention: LogCacheBlock.put() returns false when the block is full or when adding a record would cause a stream's offset delta to overflow Integer.MAX_VALUE. A new 'overflow' flag is tracked in LogCacheBlock and included in isFull(). StreamCache.add() checks the delta before accepting a record.

2. Callers handle the put() semantic (true=added, false=not added):
- S3Storage.handleAppendCallback: archives the current block and retries
   - SnapshotReadCache: archives the current block and retries
- WAL recovery path: saves rejected records as pending for the next block

3. StreamObjectCompactor.group0(): adds an offset delta guard that splits object groups when endOffset - startOffset would exceed Integer.MAX_VALUE.

4. WAL recovery refactored into standalone WALRecovery class:
   - Removed discontinuous record handling (WAL records are ordered)
   - Recovery loop with BlockHandler processes blocks incrementally
   - Pending record mechanism ensures no records are lost across blocks

Files changed:
- NEW: WALRecovery.java - extracted WAL recovery logic from S3Storage
- LogCache.java - overflow flag in LogCacheBlock, delta check in StreamCache
- S3Storage.java - handle put() returning false, use WALRecovery
- SnapshotReadCache.java - handle put() returning false
- StreamObjectCompactor.java - offset delta guard in group0()
- S3StorageTest.java - updated recovery tests for new semantics

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -46,7 +46,6 @@ import com.automq.stream.s3.wal.RecordOffset;
 import com.automq.stream.s3.wal.RecoverResult;
 import com.automq.stream.s3.wal.WriteAheadLog;
 import com.automq.stream.s3.wal.exception.OverCapacityException;
-import com.automq.stream.utils.ExceptionUtil;
 import com.automq.stream.utils.FutureTicker;
 import com.automq.stream.utils.FutureUtil;
 import com.automq.stream.utils.Systems;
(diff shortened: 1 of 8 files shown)


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3336

Gezi-lzq (approved):

PR #3350, opened 2026-05-09
fix(s3stream): avoid aliasing LogCache merge sources (#3349)

superhx:
## Summary
- Fix LogCache merge candidate construction so it copies StreamCache containers instead of aliasing source block StreamCaches.
- Prevent an aborted tryMerge swap from mutating source blocks and leaving duplicate StreamRecordBatch references to be freed twice.
- Add a regression test proving mergeBlock does not alias source StreamCache containers.

## Bug flow
1. tryMerge selects two adjacent free blocks, left and right, then releases the write lock before doing expensive merge work.
2. The previous mergeBlock(newBlock, left) put left's StreamCache object directly into the candidate newBlock when newBlock did not have that stream yet.
3. mergeBlock(newBlock, right) then appended right's records into that same StreamCache object, so candidate construction had already mutated left by alias.
4. If the final guarded swap was aborted because the source blocks changed while the lock was released, left and right stayed in the block list, but left could now contain right's record references.
5. A later free path could release the same StreamRecordBatch through both original blocks, causing the second ByteBuf release to hit refCnt 0 and throw IllegalReferenceCountException.

## Test Plan
- ./gradlew :s3stream:test --tests com.automq.stream.s3.cache.LogCacheTest --tests com.automq.stream.s3.S3StorageTest

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/cache/LogCache.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/cache/LogCache.java
@@ -413,16 +413,14 @@ public class LogCache {
         synchronized (left) {
             left.size.addAndGet(right.size());
             left.lastRecordOffset = right.lastRecordOffset;
-            left.map.forEach((streamId, leftStreamCache) -> {
-                StreamCache rightStreamCache = right.map.get(streamId);
-                if (rightStreamCache != null) {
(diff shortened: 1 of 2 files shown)


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3349

Gezi-lzq (approved):

PR #3351, opened 2026-05-09
feat(zerozone): add async partition snapshot long polling (#3344)

superhx:
## Summary
- Simplify partition snapshot long polling so each `PendingLongPoll` owns the client response future, a `triggerCf` wakeup future, and its timeout task.
- Register the pending request before snapshot collection so same-session retries reuse the in-flight future; immediate deltas still wait for partition snapshot completion before responding.
- Complete trigger/timeout wakeups through `triggerCf`, force-complete triggered long polls even when the collected delta is empty, and centralize cleanup in `pending.future.whenComplete`.

## Why
- Saves CPU on snapshot-read clients and brokers by reducing repeated empty snapshot-delta requests while keeping the pending long-poll lifecycle explicit and race-resistant.

## Test Plan
- [x] `./gradlew :core:test --tests kafka.automq.partition.snapshot.PartitionSnapshotsManagerTest`


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3344

Gezi-lzq (approved):

PR #3355, opened 2026-05-10
docs(devkit): document broker log location

superhx:
## Summary
- Document that `just logs` only shows container stdout/stderr
- Add examples for inspecting detailed broker logs under `/tmp/kafka-logs` with `just exec`
- Use full Gradle build as the default devkit build command

### DIFF
--- a/devkit/README.md
+++ b/devkit/README.md
@@ -83,6 +83,17 @@ just logs 0 | head -20          # oldest entries
 just logs 0 | tail -20          # newest entries
 ```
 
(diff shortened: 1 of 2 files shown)


1sonofqiu (approved):

Gezi-lzq (approved):
Reviewed. The README clarification about broker log locations matches devkit behavior, and the justfile build command simplification is consistent with standard local build usage. No blocking issues from my side.

PR #3370, opened 2026-05-27
fix: dedupe S3 metadata after snapshot replay

### LINKED ISSUE #3369: Fix duplicate S3 stream objects after snapshot replay on non-empty metadata image
## Description

`MetadataLoader.handleLoadSnapshot()` replays snapshot records on top of the current `MetadataImage`, and relies on each delta `finishSnapshot()` implementation to make the final image equivalent to the full snapshot state.

The AutoMQ S3 metadata deltas currently do not fully follow this contract:

1. `MetadataDelta.finishSnapshot()` does not include AutoMQ S3 metadata deltas:
   - `S3StreamsMetadataDelta`
   - `S3ObjectsDelta`
   - `KVDelta`

2. `NodeS3WALMetadataDelta.apply()` appends `S3StreamSetObject` entries to `DeltaList` instead of upserting by `objectId`.

3. `S3StreamMetadataDelta.apply()` has similar append behavior for `S3StreamObject`.

As a result, when a non-empty image already contains an SSO/SO and the loaded snapshot also contains the same `objectId`, the final image may contain duplicate objects.

## Impact

Duplicate SSO/SO entries may cause compaction or force-split tasks to build duplicate plans for the same object.

Possible failure chain:

```text
same object appears twice in metadata image
  -> compaction / force split processes it twice
  -> first request succeeds and releases/removes the source object
  -> second request processes the same source object again
  -> Data has already been released or COMPACTED_OBJECTS_NOT_FOUND
```

## Proposed Changes

1. Make SSO replay idempotent

Update `NodeS3WALMetadataDelta.apply()` to upsert `S3StreamSetObject` by `objectId` instead of appending blindly.

2. Make SO replay idempotent

Update `S3StreamMetadataDelta.apply()` to upsert `S3StreamObject` by `objectId`.

3. Complete snapshot finalization for AutoMQ metadata

Evaluate and implement `finishSnapshot()` for:

- `S3StreamsMetadataDelta`
- `S3ObjectsDelta`
- `KVDelta`

Then call them from `MetadataDelta.finishSnapshot()` so stale S3 metadata not present in the snapshot does not remain in the image.

4. Add defensive deduplication in compaction inputs

Deduplicate force-split and stream-object-compaction inputs by `objectId`, and log a warning when duplicates are detected. This should be defense-in-depth, not a replacement for metadata image invariants.

## Test Plan

Add tests for:

1. Loading a snapshot containing an existing SSO on a non-empty image does not duplicate the SSO.
2. Loading a snapshot containing an existing SO on a non-empty image does not duplicate the SO.
3. Old S3 metadata not present in the snapshot is removed after snapshot replay.
4. Incremental replay of the same `objectId` keeps only the latest object.
5. Compaction input with duplicate `objectId` is processed once and emits a warning.

Gezi-lzq:
## Summary

- Complete snapshot finalization for AutoMQ S3/KV metadata deltas so snapshot replay removes stale metadata from non-empty images.
- In snapshot-finalization mode only, rebuild S3 stream-set-object and stream-object `DeltaList` instances from the final objectId-keyed state. Normal incremental replay keeps the existing `DeltaList.copy()` append/remove path.
- Add defensive compaction input deduplication by `objectId` with warning logs.
- Add regression coverage for same-`objectId` snapshot replay, snapshot finalization, delta changed/removed sets, and compaction duplicate inputs.

Closes #3369

## Tests

- `./gradlew :metadata:test --tests 'org.apache.kafka.image.NodeRuntimeMetadataImageTest' --tests 'org.apache.kafka.image.StreamRuntimeMetadataImageTest' --tests 'org.apache.kafka.image.S3ObjectsImageTest' --tests 'org.apache.kafka.image.KVImageTest' --tests 'org.apache.kafka.image.S3StreamsMetadataImageTest'`
- `./gradlew :s3stream:test --tests 'com.automq.stream.s3.compact.CompactionManagerTest' --tests 'com.automq.stream.s3.compact.StreamObjectCompactorTest'`
- `git diff --check`

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/image/NodeS3WALMetadataDelta.java
+++ b/metadata/src/main/java/org/apache/kafka/image/NodeS3WALMetadataDelta.java
@@ -22,6 +22,9 @@ import org.apache.kafka.common.metadata.RemoveStreamSetObjectRecord;
 import org.apache.kafka.common.metadata.S3StreamSetObjectRecord;
 import org.apache.kafka.metadata.stream.S3StreamSetObject;
 
+import com.google.common.annotations.VisibleForTesting;
+
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -36,6 +39,7 @@ public class NodeS3WALMetadataDelta {
     private final Map<Long/*objectId*/, S3StreamSetObject> addedS3StreamSetObjects = new HashMap<>();
 
     private final Set<Long/*objectId*/> removedS3StreamSetObjects = new HashSet<>();
+    private boolean snapshotReplay = false;
 
     public NodeS3WALMetadataDelta(NodeS3StreamSetObjectMetadataImage image) {
         this.image = image;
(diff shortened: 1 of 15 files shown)


metadata/src/main/java/org/apache/kafka/image/NodeS3WALMetadataDelta.java
```diff
@@ -60,12 +63,30 @@ public void replay(RemoveStreamSetObjectRecord record) {
...
-        DeltaList<S3StreamSetObject> streamSetObjects = image.getObjects().copy();
-        // add all changed stream set objects
-        addedS3StreamSetObjects.forEach((id, obj) -> streamSetObjects.add(obj));
-        // remove all removed stream set objects
-        removedS3StreamSetObjects.forEach(id -> streamSetObjects.remove(obj -> Objects.equals(obj.objectId(), id)));
+        DeltaList<S3StreamSetObject> streamSetObjects;
+        if (addedS3StreamSetObjects.isEmpty() && removedS3StreamSetObjects.isEmpty()) {
+            streamSetObjects = image.getObjects();
+        } else if (snapshotReplay) {
+            Map<Long, S3StreamSetObject> objects = new LinkedHashMap<>();
```
superhx: Maybe HashMap is enough

metadata/src/main/java/org/apache/kafka/image/S3StreamMetadataDelta.java
```diff
@@ -113,9 +129,15 @@ public S3StreamMetadataImage apply() {
         DeltaList<S3StreamObject> newS3StreamObjects;
         if (changedS3StreamObjects.isEmpty() && removedS3StreamObjectIds.isEmpty()) {
             newS3StreamObjects = image.streamObjects;
+        } else if (snapshotReplay) {
+            Map<Long, S3StreamObject> objects = new LinkedHashMap<>();
```
superhx: The same as above

metadata/src/test/java/org/apache/kafka/image/NodeRuntimeMetadataImageTest.java
```diff
@@ -118,6 +118,55 @@ public void testS3Objects() {
...
+            .setOrderId(0L)
+            .setRanges(S3StreamSetObject.encode(List.of(
+                new StreamOffsetRange(STREAM0, 0L, 100L)))));
+        delta.finishSnapshot();
+
+        List<S3StreamSetObject> objects = delta.apply().getObjects().toList();
+        assertEquals(List.of(new S3StreamSetObject(0L, BROKER0, List.of(
+            new StreamOffsetRange(STREAM0, 0L, 100L)), 0L)), objects);
+    }
+
```
superhx, line 190: - Missing new scenarios
- Missing verification for Delta

metadata/src/main/java/org/apache/kafka/image/NodeS3WALMetadataDelta.java
Gezi-lzq: Changed the snapshot materialization map to HashMap. The final SSO read paths do not depend on insertion order: orderList() sorts by orderId, and direct object assertions sort by objectId in tests where needed.

metadata/src/test/java/org/apache/kafka/image/NodeRuntimeMetadataImageTest.java
Gezi-lzq: Expanded the snapshot tests to cover update + add + remove in one scenario, and added package-level delta-state assertions for the changed/removed sets. I also added similar delta-state checks for SO, S3 objects, and KV snapshot finalization so the same concern is covered across the touched deltas.

metadata/src/main/java/org/apache/kafka/image/S3StreamMetadataDelta.java
Gezi-lzq: Changed this snapshot materialization map to HashMap as well. S3StreamMetadataImage#getStreamObjects() sorts stream objects by startOffset, so map iteration order is not part of the semantics.

superhx (approved):

PR #3387, opened 2026-06-01
fix: isolate MirrorMaker KRaft cluster IDs

Gezi-lzq:
## Summary
- keep KafkaService default KRaft cluster id aligned with the Apache Kafka system-test constant
- add an explicit cluster_id override path for AutoMQ system tests
- give MirrorMaker source and target KafkaService instances distinct cluster ids so independent AutoMQ KRaft clusters do not share the S3 WAL reservation namespace
- keep isolated controller quorums on the same cluster id as their broker service and use that id for kafka-storage format / KafkaService.cluster_id()

## Root Cause
Nightly E2E run 26712668301 failed in MirrorMaker because the source and target single-node AutoMQ clusters both formatted with the fixed test cluster id I2eXt9rvSnyhct8BYmW6-w and node.id=1. AutoMQ S3 WAL reservation keys are based on clusterId/nodeId, so the concurrent clusters fenced each other. The fenced source broker force-closed storage, closed the data-plane socket, and MirrorMaker eventually timed out waiting for Resetting offset for partition.

Apache Kafka system tests use a fixed KRaft cluster id by default, which is harmless for local-disk isolated clusters. AutoMQ keeps that default and only overrides the id for MirrorMaker, where two independent clusters run concurrently against a shared object-storage namespace.

## Verification
- python3 -m py_compile tests/kafkatest/services/kafka/kafka.py tests/kafkatest/tests/core/mirror_maker_test.py
- git diff --check

### DIFF
--- a/tests/kafkatest/services/kafka/kafka.py
+++ b/tests/kafkatest/services/kafka/kafka.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import base64
 import json
 import math
 import os.path
@@ -21,7 +20,10 @@ import re
 import signal
 import time
 import subprocess
+# // AutoMQ inject start
+import base64
 import uuid
+# // AutoMQ inject end
 
 from ducktape.services.service import Service
 from ducktape.utils.util import wait_until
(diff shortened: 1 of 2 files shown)


1sonofqiu (approved):

Gezi-lzq:
https://github.com/AutoMQ/automq/actions/runs/26738274053

PR #3391, opened 2026-06-01
test: stabilize connect e2e

Gezi-lzq:
## Summary
- Stabilize two Connect E2E cases from nightly run `26712668301`.

## Why
- `test_exactly_once_source` used a 1s `read_committed` validation consumer timeout. In the failed run, the first FETCH returned committed records a few milliseconds after the consumer timed out.
- `test_pause_and_resume_sink` paused the sink while the task was RUNNING but before the sink consumer had completed assignment/reconciliation and consumed any records.

## What Changed
- Increase the Connect exactly-once validation consumer timeout from 1s to 5s while keeping `read_committed`.
- Wait for the Connect verifiable sink to consume at least one record before pause/resume.

## Verification
- `python3 -m py_compile tests/kafkatest/tests/connect/connect_distributed_test.py`
- `git diff --check`
- Local ducktape:
  - `ConnectDistributedTest.test_exactly_once_source` with `clean=false, connect_protocol=eager, metadata_quorum=ISOLATED_KRAFT, use_new_coordinator=true`: PASS, session `2026-06-01--003`, runtime `6m21s`.
  - `ConnectDistributedTest.test_pause_and_resume_sink` with `connect_protocol=compatible, metadata_quorum=ISOLATED_KRAFT, use_new_coordinator=true, group_protocol=consumer`: PASS, session `2026-06-01--002`, runtime `1m50s`.

### DIFF
--- a/tests/kafkatest/services/console_consumer.py
+++ b/tests/kafkatest/services/console_consumer.py
@@ -22,7 +22,7 @@ from ducktape.utils.util import wait_until
 from kafkatest.directory_layout.kafka_path import KafkaPathResolverMixin
 from kafkatest.services.monitor.jmx import JmxMixin, JmxTool
 from kafkatest.version import DEV_BRANCH, LATEST_0_8_2, LATEST_0_9, LATEST_0_10_0, V_0_10_0_0, V_0_11_0_0, V_2_0_0, LATEST_3_7
-from kafkatest.services.kafka.util import fix_opts_for_new_jvm, fix_snappy_for_aarch64
+from kafkatest.services.kafka.util import fix_opts_for_new_jvm
 
 """
 The console consumer is a tool that reads data from Kafka and outputs it to standard output.
@@ -65,7 +65,7 @@ class ConsoleConsumer(KafkaPathResolverMixin, JmxMixin, BackgroundThreadService)
                  enable_systest_events=False, stop_timeout_sec=35, print_timestamp=False, print_partition=False,
                  isolation_level="read_uncommitted", jaas_override_variables=None,
                  kafka_opts_override="", client_prop_file_override="", consumer_properties={},
-                 wait_until_partitions_assigned=False, log_level="DEBUG", fix_snappy_classpath=False):
+                 wait_until_partitions_assigned=False, log_level="DEBUG"):
         """
         Args:
             context:                    standard context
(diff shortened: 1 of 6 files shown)


Gezi-lzq:
Connect update after local ducktape rerun:

Scope update:

- This PR now keeps the code change scoped to Connect. The previous Streams hard-bounce wait was removed.
- Reason: the Streams failure was documented as a low-probability AutoMQ WAL failover race. We should keep that investigation record, but not change `streams_broker_bounce_test.py` in this PR just to avoid the race.

Root-cause summary:

1. `test_exactly_once_source`
   - The failed CI run did not show a transaction visibility/product bug.
   - The debug consumer with `read_uncommitted` saw records, and the failed `read_committed` consumer sent its first FETCH at `13:32:44.957`, timed out at `13:32:44.961`, then the broker returned non-empty records at `13:32:45.006` with `highWatermark=4496` and `lastStableOffset=4496`.
   - The 1s console-consumer timeout can be consumed by group join, offset lookup, list offsets, and first fetch latency before any record is delivered to the process.
   - AutoMQ makes this more visible because the system test runs KRaft + S3/WAL + new coordinator in a containerized CI environment, so startup/recovery and first-poll latency are higher than the upstream Kafka local-disk path.
   - The fix keeps `read_committed`; it only changes the validation timeout from 1s to 5s. This gives the validator a small buffer for the first committed fetch, while still failing quickly enough to expose real stalls.

2. `test_pause_and_resume_sink`
   - REST `RUNNING` only means the Connect task has started. It does not prove the sink consumer has completed assignment/reconciliation or consumed any record.
   - In the failed run, the sink task entered the new consumer-group protocol `RECONCILING` state and the test paused the connector before any verifiable sink output was emitted.
   - With AutoMQ KRaft + S3/WAL + new coordinator + consumer group protocol, assignment/reconcile/offset-reset timing is slower and the test race becomes observable.
   - The fix waits for at least one sink record before pause, so the pause/resume assertion starts from a real consumption baseline instead of a task-started-only state.

Local verification:

- `ConnectDistributedTest.test_exactly_once_source` with `clean=false, connect_protocol=eager, metadata_quorum=ISOLATED_KRAFT, use_new_coordinator=true`: PASS locally after reducing the validation timeout to 5s, ducktape session `2026-06-01--003`, runtime `6m21s`.
- `ConnectDistributedTest.test_pause_and_resume_sink` with `connect_protocol=compatible, metadata_quorum=ISOLATED_KRAFT, use_new_coordinator=true, group_protocol=consumer`: PASS locally, ducktape session `2026-06-01--002`, runtime `1m50s`.

Local test note: the first local rerun reached validation/teardown but failed while copying service logs because the host mount had only ~8GB free. After freeing space and rebuilding ducker nodes, both exact failed matrices passed.

1sonofqiu (dismissed):

1sonofqiu (approved):

PR #3414, opened 2026-06-10
fix(s3stream): validate object WAL reservation before acquire write

Gezi-lzq:
## Summary

This patch changes object WAL reservation acquire from blind write to read-then-write validation.

Before writing the reservation object, `ObjectReservationService` now reads the existing reservation and applies the reservation state machine:

- normal acquire can create a missing reservation
- failover acquire cannot create a missing reservation
- repeated acquire with the same normal reservation is idempotent
- failover can acquire the same epoch only from an existing normal reservation
- normal acquire cannot overwrite a same-epoch failover reservation
- higher epoch normal acquire can replace an older reservation

The write path still uses the existing object storage `write` API. This change does not introduce conditional write, ETag metadata, or object storage API changes.

## Motivation

The previous implementation wrote the reservation object directly during acquire. That allowed a restarted node to blindly overwrite a failover reservation with the same epoch, which could break the intended ownership protection during failover/recover races.

By reading and validating the current reservation before writing, acquire preserves the reservation state machine without requiring backend conditional-write support.

## Compatibility and Correctness

This patch intentionally keeps acquire on the plain write path for object-storage compatibility. Some S3-compatible backends either do not support `If-Match` / `If-None-Match` on `PutObject`, or their behavior is not documented consistently enough to rely on it in the common object storage layer.

A concurrent change after the read may still overwrite this write, because this patch does not provide storage-level CAS. This does not make the caller observe a valid reservation unless the final object content still matches the requested node/epoch/failover state. The acquire state machine prevents known invalid transitions, and the existing verification path remains responsible for checking the final reservation content.

## Changes

- Add read-before-write validation in `ObjectReservationService.acquire`
- Parse and validate existing reservation object content before deciding whether to write
- Reject invalid acquire transitions instead of blindly overwriting
- Keep object storage writes on the existing `write` path
- Add unit coverage for acquire state transitions

## Tests

```bash
./gradlew :s3stream:test --tests com.automq.stream.s3.wal.impl.object.ObjectReservationServiceTest
```

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/wal/impl/object/ObjectReservationService.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/wal/impl/object/ObjectReservationService.java
@@ -21,9 +21,11 @@ package com.automq.stream.s3.wal.impl.object;
 
 import com.automq.stream.s3.ByteBufAlloc;
 import com.automq.stream.s3.Constants;
+import com.automq.stream.s3.exceptions.ObjectNotExistException;
 import com.automq.stream.s3.network.ThrottleStrategy;
 import com.automq.stream.s3.operator.ObjectStorage;
 import com.automq.stream.s3.wal.ReservationService;
+import com.automq.stream.utils.FutureUtil;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


s3stream/src/main/java/com/automq/stream/s3/wal/impl/object/ObjectReservationService.java
```diff
@@ -97,12 +99,145 @@ public CompletableFuture<Boolean> verify(long nodeId, long epoch, boolean failov
...
+        if (current.nodeId() != nodeId) {
+            LOGGER.warn("Overwrite reservation object with mismatched node id, expect={}, actual={}", nodeId, current.nodeId());
+            return AcquireDecision.write();
+        }
+
+        // Existing reservation transition table:
+        // - failover acquire: same-epoch normal -> WRITE, same-epoch failover -> NOOP, otherwise REJECT.
+        // - normal acquire: same-epoch normal -> NOOP, newer epoch -> WRITE, otherwise REJECT.
+        if (failover) {
+            if (current.epoch() == epoch) {
```
superhx: Use current.epoch() <= epoch

superhx (approved):

PR #3415, opened 2026-06-10
fix(shell): preserve '=' in S3Url credentials when parsing

karanray06:
S3Url.parse split each query parameter on every '=' via split("="), so access/secret keys that contain '=' (common in base64-encoded credentials, including '=' padding) were either rejected with "Invalid parameter format" or silently truncated. Split on the first '=' only via split("=", 2) to preserve the full value.

comment to ping reviewers.*


superhx:
@karanray06 karanray06 It is expected that the values of the URL parameters are processed by URLEncoder. Theoretically, special characters such as the equals sign (=) will be escaped.

PR #3425, opened 2026-06-17
fix(s3stream): limit cleanup v1 compaction scope (#3423)

superhx:
- make CLEANUP_V1 bypass general grouping and only compact the first dirty composite object after expired cleanup
- remove obsolete CLEANUP_V1 branches from shared compaction filter/group validation paths
- trigger MAJOR_V1 at 90% of the object-count threshold while preserving small-normal-object skipping before the hard threshold


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3423

Gezi-lzq (approved):

PR #3428, opened 2026-06-17
fix(log): release records after offset metadata lookup (#3427)

superhx:
- release pooled records created during LocalLog offset metadata lookup
- add regression coverage for releasing records returned by the internal read


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3427

Gezi-lzq (approved):

PR #3431, opened 2026-06-19
test: cover OTLP gRPC metrics exporter

### LINKED ISSUE #3217: [BUG] OTLP gRPC metrics exporter crashes on startup — GrpcSenderProvider missing from classpath

ns3154:
## Summary
- Add an S3Unit regression test for the OTLP gRPC metrics exporter.
- Build the metric reader with protocol=grpc so missing OpenTelemetry gRPC sender SPI fails during tests instead of broker startup.

Closes #3217

## Tests
- `JAVA_HOME=<jdk17-home> PATH=<jdk17-home>/bin:$PATH ./gradlew --no-daemon :automq-metrics:S3UnitTest --tests com.automq.opentelemetry.exporter.OTLPMetricsExporterTest`
- `JAVA_HOME=<jdk17-home> PATH=<jdk17-home>/bin:$PATH ./gradlew --no-daemon --build-cache rat checkstyleMain checkstyleTest spotlessJavaCheck`

### DIFF
--- a/automq-metrics/src/test/java/com/automq/opentelemetry/exporter/OTLPMetricsExporterTest.java
+++ b/automq-metrics/src/test/java/com/automq/opentelemetry/exporter/OTLPMetricsExporterTest.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2026, AutoMQ HK Limited.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


Gezi-lzq:
Thanks for adding the regression coverage. The related issue has already been addressed, so this PR is no longer needed. Closing it. We appreciate your contribution.

PR #3406, opened 2026-06-08
perf(failover): accelerate failover (#3398)

superhx:

PR #3443, opened 2026-07-02
perf(failover): optimize failover to reduce latency. (#3412)

superhx:
The failover detection and recovery path is too slow, caused by three compounding issues:
* Polling-only detection with high latency The FailoverControlManager relies a periodic scheduled task running every 1 second to detect fenced brokers and trigger failover. There is no event-driven mechanism — when a broker is fenced by the controller, the failover system doesn't know about it until the next polling cycle. #3406 that it attempts to compensate for detection latency with high-frequency polling, but the polling itself is blocking and consumes the controller's shared event queue — which in turn slows down all operations, including the failover task itself. The correct approach, as this optimization does, is to replace high-frequency polling with event-driven triggering and reduce the periodic scan to a low-frequency safety net (60 seconds).

* Sequential I/O during WAL recovery In `S3Storage.recover0()`, `streamManager.getOpeningStreams()` (a remote call) and `deltaWAL.recover()`, `deltaWAL.reset` and `closeStreams` can be executed in parallel instead of sequentially.

* Batch delay in FAILOVER WAL write mode During failover recovery, the only write to the WAL is a fake record from `trim()` to persist the trimOffset. In `DefaultWriter`, this record is subject to the normal batching logic (default batchInterval = 250ms), meaning the trim-offset-persisting record sits in the batch for up to 250ms before being uploaded. This delay directly adds to the total failover recovery time.

### DIFF
--- a/core/src/main/java/kafka/automq/controller/DefaultQuorumControllerExtension.java
+++ b/core/src/main/java/kafka/automq/controller/DefaultQuorumControllerExtension.java
@@ -74,4 +74,9 @@ public class DefaultQuorumControllerExtension implements QuorumControllerExtensi
             BrokerHeartbeatRequestData request, long registerBrokerRecordOffset) {
         return Optional.empty();
     }
+
+    @Override
+    public void onBrokerFenced(int brokerId) {
+        failoverControlManager.triggerFailover();
+    }
 }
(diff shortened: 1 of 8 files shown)


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3412

Gezi-lzq (approved):

PR #3449, opened 2026-07-03
feat(tabletopic): derive identifiers from Debezium key

### LINKED ISSUE #3448: Support deriving TableTopic identifier fields from Debezium Kafka key
## Background

This capability is only intended for Debezium CDC scenarios. Debezium change event keys use the primary key columns of the source database table, so the Kafka key schema naturally contains the identifier column names.

References:

- Debezium PostgreSQL change event key: https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-change-events-key
- Debezium topic routing unique key: https://debezium.io/documentation/reference/stable/transformations/topic-routing.html#_ensure_unique_key

Today, TableTopic requires users to explicitly configure `automq.table.topic.id.columns`. This becomes costly to manage when many Debezium tables use different primary key column names.

## Goal

Support deriving identifier columns from the Kafka key schema in Debezium scenarios:

```properties
automq.table.topic.id.columns=[_from_debezium_key_]
```

The resolved columns are synced to Iceberg table identifier fields and used for CDC/upsert writes.

## Plan

- Support `[_from_debezium_key_]` as a special `id.columns` value.
- Reject mixed usage such as `[_from_debezium_key_, id]`.
- Support this only for Debezium records when the key uses a schema-based converter.
- Resolve identifier column names from the Kafka key schema.
- Fail clearly if the key schema is missing or is not a record.
- Sync non-empty identifier columns to Iceberg table metadata.
- If existing Iceberg identifier fields differ, overwrite them and log previous/new fields.
- If `id.columns` is empty, leave existing Iceberg identifier fields unchanged.
- Make writers use `table.schema().identifierFieldIds()` instead of reparsing config during writes.

Gezi-lzq:
## Summary

- Add `automq.table.topic.id.columns=[_from_debezium_key_]` to derive Table Topic identifier columns from Debezium key schema.
- Propagate resolved identifier columns through schema identity, `ProcessingResult`, Iceberg schema metadata, and equality-delete writes.
- Require `_from_debezium_key_` to use `automq.table.topic.convert.key.type=by_schema_id` at server-side topic config validation.
- Keep runtime null/missing Debezium keys silent: they resolve to empty identifiers.

Fixes #3448

## Notes

- `_from_debezium_key_` must be used alone in `automq.table.topic.id.columns`.
- Key fields must exist as top-level fields after `flatten_debezium`; otherwise schema binding fails clearly.
- DevKit/KRaft create and alter validation paths are covered. ZK admin path coverage is a remaining boundary if TableTopic still supports ZK mode.

## Verification

- Focused unit tests for processor, binder, Iceberg table manager, and config validation.
- Required local gate:

```bash
./gradlew --build-cache rat checkstyleMain checkstyleTest spotlessJavaCheck
```

- DevKit regression with Debezium + Schema Registry + Iceberg REST:
  - stable PK derives Iceberg `identifier-field-ids` and writes equality deletes
  - null key / no key schema is silent and creates an identifier-less table schema
  - explicit identifier config can be changed to `_from_debezium_key_`, updating Iceberg identifier metadata
  - invalid `_from_debezium_key_` + non-`by_schema_id` key converter is rejected on create and alter

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/config/TopicConfig.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/TopicConfig.java
@@ -310,8 +310,8 @@ public class TopicConfig {
         "Note: 'flatten_debezium' requires schema-based conversion.";
 
     public static final String TABLE_TOPIC_ID_COLUMNS_CONFIG = "automq.table.topic.id.columns";
-    public static final String TABLE_TOPIC_ID_COLUMNS_DOC = "The primary key, comma-separated list of columns that identify a row in tables."
-        + "ex. [region, name]";
+    public static final String TABLE_TOPIC_ID_COLUMNS_DOC = "The primary key, comma-separated list of columns that identify a row in tables. "
+        + "ex. [region, name]. Use [_from_debezium_key_] to derive identifier columns from the schema-based Debezium Kafka key.";
     public static final String TABLE_TOPIC_PARTITION_BY_CONFIG = "automq.table.topic.partition.by";
     public static final String TABLE_TOPIC_PARTITION_BY_DOC = "The partition fields of the table. ex. [bucket(name), month(timestamp)]";
     public static final String TABLE_TOPIC_UPSERT_ENABLE_CONFIG = "automq.table.topic.upsert.enable";
--- a/core/src/main/java/kafka/automq/table/binder/RecordBinder.java
+++ b/core/src/main/java/kafka/automq/table/binder/RecordBinder.java
@@ -65,6 +65,14 @@ public class RecordBinder {
         this(AvroSchemaUtil.toIceberg(avroRecord.getSchema()), avroRecord.getSchema());
     }
 
+    /**
+     * Creates a binder whose Iceberg schema carries identifier field IDs resolved from column names.
+     *
+     * @param avroRecord source Avro record used to derive the Iceberg schema
+     * @param identifierColumns identifier column names to resolve, or empty when the schema has no identifiers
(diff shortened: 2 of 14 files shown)


Gezi-lzq:
I ran a DevKit pass against the current implementation and wrote the experiment report here:

`/home/debian/automq-workspace/scratch/tabletopic-from-debezium-key-devkit-report.zh.md`

Current strategic conclusion:

- The core path is valid: stable Debezium PK, Avro key schema, `convert.key.type=by_schema_id`, and `flatten_debezium` can derive Iceberg identifier fields from the Debezium key.
- Composite PK works.
- Existing Iceberg identifier fields can be updated to match the currently configured/derived identifier set.
- CDC update/delete continues to write equality deletes through `table.schema().identifierFieldIds()`.

The main merge risk is usability, not the happy-path correctness. `[_from_debezium_key_]` is easy to misconfigure if the key converter is not schema-aware. In that case the failure happens late in the CDC writer, which is hard to diagnose.

Recommended next steps before merging:

1. Validate that `automq.table.topic.id.columns=[_from_debezium_key_]` requires `automq.table.topic.convert.key.type=by_schema_id`.
2. Update docs/examples to always include the key converter config when using `_from_debezium_key_`.
3. Improve error messages for:
   - null key / no source PK
   - key field not present as a top-level flattened value field
4. Add focused tests for these boundary cases.

I would keep PK changes, no-PK append-only fallback, and deriving identifiers from nested `_kafka_key` out of this PR. Those are separate migration/product-design topics.

superhx (approved):

PR #3460, opened 2026-07-13
fix: add kafka-go metadata compatibility mode

superhx:
## Why

This issue was verified against kafka-go `v0.4.51`.

A default `NewWriter` refreshes metadata every 15 seconds. Produce retries up to 10 attempts with quadratic backoff from 100 ms to 1 second and a default 10-second write timeout. However, when Metadata returns `LEADER_NOT_AVAILABLE`, kafka-go converts the missing leader to a local non-temporary no-leader error, which can stop Produce before the partition becomes available.

During an AutoMQ partition reassignment, Metadata may briefly have no leader while `replicas[0]` is already the broker opening the partition. This timing can cause user-visible write failures.

## What

Add a kafka-go Metadata compatibility mode, enabled by default through:

```properties
automq.kafka-go.metadata.compatibility.enabled=true
```

For likely kafka-go clients, AutoMQ rewrites partition-level `LEADER_NOT_AVAILABLE` to `NONE` and uses `replicas[0]` as the predictive leader when replicas are non-empty. The setting is dynamically configurable.

This does not change kafka-go metadata refresh timing. Users should set `WriterConfig.RebalanceInterval` to 5 seconds to reduce stale routing during reassignment.

## Validation

Three-node devkit, 10 msg/s, reassignment every 30 seconds for 5 minutes, kafka-go `v0.4.51`, `RebalanceInterval=5s`:

- compatibility disabled: 236 no-leader errors
- compatibility enabled: 0 errors

### DIFF
--- a/core/src/main/java/kafka/automq/AutoMQConfig.java
+++ b/core/src/main/java/kafka/automq/AutoMQConfig.java
@@ -43,6 +43,7 @@ import java.util.concurrent.TimeUnit;
 import static org.apache.kafka.common.config.ConfigDef.Importance.HIGH;
 import static org.apache.kafka.common.config.ConfigDef.Importance.LOW;
 import static org.apache.kafka.common.config.ConfigDef.Importance.MEDIUM;
+import static org.apache.kafka.common.config.ConfigDef.Range.atLeast;
 import static org.apache.kafka.common.config.ConfigDef.Range.between;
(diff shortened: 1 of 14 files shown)


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3459

Gezi-lzq (approved):

PR #3463, opened 2026-07-17
feat(s3stream): recycle sequential byte buffer slabs

superhx:
## Summary

Introduce `RecyclingByteBufSeqAlloc` to retain and reuse sequential allocation
slabs instead of repeatedly allocating and releasing large backing buffers.

This change also makes record encode/decode allocator ownership explicit and
migrates applicable `ByteBufSeqAlloc` users to the recycling allocator.

## Motivation

With `POOLED_HEAP`, Netty's 4 MiB heap chunk is backed by a `byte[]`. After the
array header and object alignment are included, the object exceeds 4 MiB and
occupies 6 MiB of ZGC large-page capacity.

The previous sequential allocator also released its backing buffer after all
retained slices were released. Under sustained append and batched LogCache
eviction, this caused large buffers to repeatedly enter old generation and wait
for collection instead of being reused.

## Implementation

- Add **`RecyclingByteBufSeqAlloc`** with:
  - sequential retained-slice allocation;
  - retired slab detection through owner `refCnt`;
  - LIFO reuse of free slabs;
  - one-minute idle eviction;
  - periodic maintenance for low-traffic workloads;
  - heap and direct allocation support;
  - thread-safe allocation and close behavior.
- Set heap slab payload to `4 MiB - 64 B`, keeping the backing array within
  4 MiB of ZGC page capacity.
- Source direct slabs from `ByteBufAlloc` so they continue using Netty's pooled
  allocator.
- Scale allocator concurrency with:
  `min(CPU, max(heap size / 2 GiB, 1))`.
- Export retained slab memory through
  `kafka_stream_buffer_allocated_memory_size` with `source="seq_alloc"`.
- Add the `source` label to existing `ByteBufAlloc` metric series.
- Replace applicable sequential allocators in stream append, WAL, snapshot,
  compaction, link-record, and block-cache paths.
- Require explicit allocators for `StreamRecordBatch` encode/decode operations.
- Distinguish copied and retained `ObjectReader.DataBlockGroup` iteration.
- Classify copied block-cache records as `BLOCK_CACHE`.
- Add allocator unit coverage and include affected AutoMQ tests in `S3Unit`.

Process-wide allocators intentionally remain alive for the process lifetime so
unrelated instances of the same component can reuse a bounded active slab set.

## Compatibility

- No configuration changes are required.
- The allocator mode follows the process-wide `ByteBufAlloc` policy selected
  during startup.
- Existing allocated-memory metrics gain a `source="byte_buf_alloc"` label.
  Recycling allocator series use `source="seq_alloc"`.

### DIFF
--- a/core/src/main/java/kafka/automq/partition/snapshot/ConfirmWalDataDelta.java
+++ b/core/src/main/java/kafka/automq/partition/snapshot/ConfirmWalDataDelta.java
@@ -198,10 +198,14 @@ public class ConfirmWalDataDelta implements ConfirmWAL.AppendListener {
         }
         List<StreamRecordBatch> records = new ArrayList<>();
         ByteBuf buf = Unpooled.wrappedBuffer(data);
-        while (buf.readableBytes() > 0) {
-            StreamRecordBatch record = StreamRecordBatch.parse(buf, false);
-            records.add(record);
+        try {
+            while (buf.readableBytes() > 0) {
+                StreamRecordBatch record = StreamRecordBatch.parse(buf, false, null);
+                records.add(record);
+            }
+            return records;
+        } finally {
+            buf.release();
         }
-        return records;
     }
 }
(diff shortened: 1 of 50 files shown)


superhx:
This PR cherry-picks #3461 into the `main` branch.

Gezi-lzq (approved):

PR #3465, opened 2026-07-17
fix: broker cannot gracefully shutdown when fenced by a duplicate node

### LINKED ISSUE #718: [BUG] Broker cannot gracefully shutdown when a duplicated broker attempts to start
### What went wrong?

When a broker with duplicated node id attempts to start (will not succeed), the existing broker will not be able to gracefully close its streams
image

### How to reproduce the issue?

1. start a broker with node id 1
2. create some topics and send some messages
3. start another broker with node id 1 (start failed)
4. the original broker with node id 1 will not be able to shutdown gracefully

### Additional information

Please attach any relevant logs, backtraces, or metric charts.

vishalmaurya850:
Fixes #718

Fixes a shutdown hang risk by ensuring WAL-append failures propagate to the caller, rather than leaving append futures incomplete and potentially blocking stream close paths.

The previous implementation propagated the WAL append failure only through the aggregate future, while individual append futures remained incomplete. As a result, components waiting on those futures (such as the stream close path) could block indefinitely if the WAL append failed.

Changes:

- Complete the per-append future exceptionally when `deltaWAL.append(...)` completes exceptionally, ensuring callers observe the failure immediately instead of waiting indefinitely.
- Preserve existing successful append behavior while guaranteeing failed WAL appends always unblock dependent operations, including stream shutdown.

*Testing strategy:*

- Existing `S3StreamClient` and `S3Stream` unit tests continue to pass, confirming no regression in the normal append and shutdown flows.
- The change is limited to the WAL append failure path and ensures append futures are completed exceptionally whenever the underlying WAL append fails, preventing indefinite waits during stream shutdown.
- Verified that successful append behavior remains unchanged while failure scenarios now propagate correctly to callers.

- [X] Verify design and implementation
- [X] Verify test coverage and CI build status
- [X] Verify documentation (including upgrade notes)

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -405,7 +405,9 @@ public class S3Storage implements Storage {
         }).whenComplete((nil, ex) -> {
             if (ex != null) {
                 LOGGER.error("append WAL fail", ex);
-                request.cf.completeExceptionally(ex);
+                Throwable cause = FutureUtil.cause(ex);
+                EventLoop executor = callbackExecutors[Math.abs((int) (request.record.getStreamId() % callbackExecutors.length))];
+                executor.execute(() -> request.cf.completeExceptionally(cause));
                 storageFailureHandler.handle(ex);
             }
         });
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -405,15 +405,15 @@ public class S3Storage implements Storage {
         }).whenComplete((nil, ex) -> {
             if (ex != null) {
                 LOGGER.error("append WAL fail", ex);
-Throwable cause = FutureUtil.cause(ex);
-EventLoop executor = callbackExecutors[Math.abs((int) (request.record.getStreamId() % callbackExecutors.length))];
-try {
-    executor.execute(() -> request.cf.completeExceptionally(cause));
-} catch (IllegalStateException ignored) {
-    // Callback executor might already be shut down during broker/stream shutdown; complete directly to avoid hanging callers.
-    request.cf.completeExceptionally(cause);
-}
-storageFailureHandler.handle(ex);
+                Throwable cause = FutureUtil.cause(ex);
+                EventLoop executor = callbackExecutors[Math.abs((int) (request.record.getStreamId() % callbackExecutors.length))];
+                try {
+                    executor.execute(() -> request.cf.completeExceptionally(cause));
+                } catch (IllegalStateException ignored) {
+                    // Callback executor might already be shut down during broker/stream shutdown; complete directly to avoid hanging callers.
+                    request.cf.completeExceptionally(cause);
+                }
+                storageFailureHandler.handle(ex);
             }
         });
         return false;
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -405,8 +405,8 @@ public class S3Storage implements Storage {
         }).whenComplete((nil, ex) -> {
             if (ex != null) {
                 LOGGER.error("append WAL fail", ex);
+                request.cf.completeExceptionally(ex);
                 storageFailureHandler.handle(ex);
-                return;
             }
         });
         return false;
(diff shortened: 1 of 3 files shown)


superhx:
@vishalmaurya850  May I ask what scenario caused the duplicate nodeId? Theoretically, nodeId should not be duplicated during deployment.

vishalmaurya850:
@superhx Thanks for having a look on my PR

Most common cause: Kubernetes pod restart races — the new pod starts before the old one fully terminates, briefly creating two brokers with the same node ID.

Also: operator config mistakes, VM clones, or rolling upgrade timing issues.

It's an edge case, but the broker shouldn't hang forever when it happens.

superhx:
@vishalmaurya850 Judging from the problem description, can it be shut down successfully as long as this issue 'Pending WAL appends never complete (WAL is fenced)' is resolved?

vishalmaurya850:
@superhx Yes, largely. The main blocker is the hanging pending WAL appends. Once those complete or fail fast, StreamWrapper.close() removes the stream from tracking even if closeStream returns NODE_EPOCH_EXPIRED, so shutdown proceeds. The NODE_EPOCH_EXPIRED handling is just for cleaner logs, and the force-close timeout is the safety net to guarantee appends don't hang.

superhx:
> @superhx Yes, largely. The main blocker is the hanging pending WAL appends. Once those complete or fail fast, StreamWrapper.close() removes the stream from tracking even if closeStream returns NODE_EPOCH_EXPIRED, so shutdown proceeds. The NODE_EPOCH_EXPIRED handling is just for cleaner logs, and the force-close timeout is the safety net to guarantee appends don't hang.

Could we adjust this to fix the issue where Pending WAL appends never complete? Right now it mostly acts as a fallback measure.

vishalmaurya850:
@superhx Let me deep dive in the code and will let you know If we can solve it.

vishalmaurya850:
Complete request.cf exceptionally when the WAL append fails, so pending appends drain on fencing instead of blocking shutdown indefinitely.

vishalmaurya850:
@superhx I have created the fix

superhx:
> @superhx I have created the fix

Great, after this fix, do we still need to modify the existing S3Stream and S3StreamClient?

vishalmaurya850:
@superhx The S3Storage fix is necessary but not sufficient — the S3Stream and S3StreamClient changes remain essential for shutdown resilience.

superhx:
> @superhx The S3Storage fix is necessary but not sufficient — the S3Stream and S3StreamClient changes remain essential for shutdown resilience.

Could you please describe why these fixes are necessary?

vishalmaurya850:
The S3Storage fix only covers the append path — it doesn't touch the close path, which has its own independent failure points:

- S3Stream: close0() calls forceUpload(streamId) and closeStream() — both can fail with fencing errors during shutdown. Without handling these, the close future fails and the stream gets stuck.

- S3StreamClient: The S3Storage fix intentionally leaves the caller's future incomplete, which means pending appends never resolve → close() waits on them forever → the infinite for(;;) loop in closeStreams() never breaks. The timeout + force-close escalation is what unblocks this.

superhx:
> The S3Storage fix only covers the append path — it doesn't touch the close path, which has its own independent failure points:
> 
> * S3Stream: close0() calls forceUpload(streamId) and closeStream() — both can fail with fencing errors during shutdown. Without handling these, the close future fails and the stream gets stuck.
> * S3StreamClient: The S3Storage fix intentionally leaves the caller's future incomplete, which means pending appends never resolve → close() waits on them forever → the infinite for(;;) loop in closeStreams() never breaks. The timeout + force-close escalation is what unblocks this.

If we fix the append stuck issue, close should theoretically not get stuck.

vishalmaurya850:
> If we fix the append stuck issue, close should theoretically not get stuck.

You are right. We can remove it if you prefer; however, I believe it should remain as a fallback mechanism to ensure that shutdown is handled gracefully under all circumstances.

superhx:
> > If we fix the append stuck issue, close should theoretically not get stuck.
> 
> You are right. We can remove it if you prefer; however, I believe it should remain as a fallback mechanism to ensure that shutdown is handled gracefully under all circumstances.

Please remove it in case it does not turn out to be the root cause.

vishalmaurya850:
> > > If we fix the append stuck issue, close should theoretically not get stuck.
> > 
> > 
> > You are right. We can remove it if you prefer; however, I believe it should remain as a fallback mechanism to ensure that shutdown is handled gracefully under all circumstances.
> 
> Please remove it in case it does not turn out to be the root cause.

I have took back the commit for both S3Stream, S3StreamClient

superhx (approved):

PR #3466, opened 2026-07-19
fix: change log.dirs and s3.wal.path defaults away from /tmp

### LINKED ISSUE #683: change log.dirs and s3.wal.path default value when using admin tool
Apache Kafka use /tmp by default. Tmp directory use tmpfs in many operating system that may cause unexpected swap behaviour that may cause performance issue. Use other directory to avoid the situation.

pradhyumanarora:
## Summary

Fixes #683

Defaulting log.dirs and s3.wal.path to paths under /tmp is dangerous in production environments where /tmp is memory-backed (e.g., 	mpfs on Linux). Data written there is lost on reboot. This PR changes the defaults to /var/kafka/, which is the conventional stable directory for Kafka data on Linux.

## Changes

| File | Old default | New default |
|------|-------------|-------------|
| config/kraft/broker.properties | /tmp/kraft-broker-logs | /var/kafka/kraft-broker-logs |
| config/kraft/controller.properties | /tmp/kraft-controller-logs | /var/kafka/kraft-controller-logs |
| config/kraft/server.properties | /tmp/kraft-combined-logs | /var/kafka/kraft-combined-logs |
| server-common/.../ServerLogConfigs.java | /tmp/kafka-logs | /var/kafka/logs |
| s3stream/.../Config.java |

### DIFF
--- a/config/kraft/broker.properties
+++ b/config/kraft/broker.properties
@@ -72,7 +72,7 @@ socket.request.max.bytes=104857600
 ############################# Log Basics #############################
 
 # A comma separated list of directories under which to store log files
-log.dirs=/tmp/kraft-broker-logs
+log.dirs=/var/kafka/kraft-broker-logs
 
 # The default number of log partitions per topic. More partitions allow greater
 # parallelism for consumption, but this will also result in more files across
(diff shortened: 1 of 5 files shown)


KaimingWan:
@pradhyumanarora Hi, thank you for the contribution. We plan to continue to use the dir to ensure the config can align with Apache Kafka's default configuration. Maybe we can add some tips in the documentation to avoid misunderstanding.

PR #3470, opened 2026-07-22
perf(zerozone): improve high-throughput routing

superhx:
## Motivation

ZeroZone routing can develop reciprocal backpressure under high bidirectional traffic because RouterIn and RouterOut acquire byte permits on Kafka request-handler threads. Once one peer slows down, blocked request handlers can prevent the reciprocal router requests required to release permits.

The routing path also has two additional sources of tail latency:

- RouterIn uses a global read-completion queue, so one slow RouterChannel read can block unrelated producers.
- A normal ElasticLog segment roll waits synchronously for segment metadata persistence before allowing append processing to continue.

## Changes

- Remove the blocking RouterIn and RouterOut permit limiters from the request path.
- Reuse `InterBrokerAsyncSender` for broker-to-broker request delivery.
- Add a bounded, heap-scaled, one-shot cache to `ObjectRouterChannel` so recently appended remote payloads can be consumed without reading them back from object storage.
- Replace the RouterIn global completion queue with fixed striped order queues:
  - preserve arrival order for the same `orderHint`;
  - allow independent ordering stripes to make progress concurrently;
  - use `min(65536, CPU cores * 64)` stripes;
  - map the complete unsigned 16-bit `orderHint` space consistently.
- Add detailed ZeroZone router latency stages:
  - `out/handle_request`
  - `out/append_channel`
  - `out/request_target`
  - `in/handle_request`
  - `in/get_channel`
  - `in/append_partition`
- Avoid blocking normal `ElasticLog.roll()` on segment metadata acknowledgement. Segment metadata and subsequent appends remain ordered by the same WAL, while asynchronous persistence failures continue to be reported.

## Correctness

Requests with the same `orderHint` are still submitted to the append EventLoop in receive order, preserving Kafka producer sequence ordering. A failed RouterChannel read or scheduling attempt advances its ordering stripe instead of blocking later requests indefinitely.

The ObjectRouterChannel cache transfers payload ownership exactly once: either to the first reader or to the removal path on expiration, eviction, or close.

This change does not alter the ZeroZone request protocol or require new configuration.

### DIFF
--- a/core/src/main/java/kafka/automq/zerozone/AsyncSender.java
+++ b/core/src/main/java/kafka/automq/zerozone/AsyncSender.java
@@ -22,13 +22,10 @@ package kafka.automq.zerozone;
 import kafka.server.KafkaConfig;
 
 import org.apache.kafka.clients.ApiVersions;
-import org.apache.kafka.clients.ClientRequest;
 import org.apache.kafka.clients.ClientResponse;
 import org.apache.kafka.clients.ManualMetadataUpdater;
 import org.apache.kafka.clients.MetadataRecoveryStrategy;
 import org.apache.kafka.clients.NetworkClient;
-import org.apache.kafka.clients.NetworkClientUtils;
-import org.apache.kafka.clients.RequestCompletionHandler;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.metrics.Metrics;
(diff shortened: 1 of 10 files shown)


superhx:
Cherry-pick of #3468.

Gezi-lzq (approved):

PR #3472, opened 2026-07-22
perf(jvm): enable generational ZGC on JDK 21

superhx:
## Summary

- select ZGC collection interval options based on the runtime JDK version
- enable Generational ZGC with a five-second minor collection interval on JDK 21 and 22
- upgrade the DevKit broker runtime to JDK 21 and use heap-backed allocation without a direct-memory limit
- version the DevKit image by JDK and document the broker heap override

### DIFF
--- a/bin/kafka-server-start.sh
+++ b/bin/kafka-server-start.sh
@@ -30,6 +30,18 @@ calculate_max_heap_kb() {
     fi
 }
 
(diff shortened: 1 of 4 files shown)


superhx:
Cherry-pick of #3471.

1sonofqiu (approved):

PR #3476, opened 2026-07-24
fix(s3stream): improve V1 stream object compaction convergence

superhx:
## Summary

Improve V1 Stream Object compaction convergence and prevent small Normal Objects from permanently fragmenting Composite Objects.

## Problem

MINOR_V1 previously treated 4 MiB as a hard upper bound, while MAJOR_V1 excluded Normal Objects smaller than 4 MiB. Objects near this boundary could therefore remain too small for MAJOR_V1 and accumulate over time.

A small Normal Object between two continuous Composite Objects could also prevent the Composite Objects from being merged.

Under cluster object-count pressure, MAJOR_V1 took priority over regular V1 compaction, potentially starving MINOR_V1 while small Normal Objects were still excluded.

## Changes

- Allow MINOR_V1 to merge continuous small Normal Objects until the result reaches or crosses the 4 MiB threshold.
- Allow MAJOR_V1 to include a small Normal Object when it continuously bridges two Composite Objects.
- Keep small Normal Objects excluded from regular MAJOR_V1 compaction unless they are required to preserve continuity.
- Run object-count-pressure MAJOR_V1 after the regularly scheduled V1 compaction instead of replacing it.
- Avoid running MAJOR_V1 twice when it is already the regularly scheduled compaction.


superhx:
This PR cherry-picks the change from #3473, which has already been merged into the `1.7` branch, and applies the same change to `main`.

Gezi-lzq (approved):

PR #3485, opened 2026-07-31
fix(s3stream): handle WAL prepare failures

utkarshmankad:
## Summary
- On WAL object prepare failure, fail the upload future and invoke the storage failure handler instead of silently swallowing the exception.
- Fix a resulting deadlock: the failed context was never polled off `walPrepareQueue`, so the prepare pipeline stayed permanently non-empty and never accepted new prepare cycles once a failure occurred (only masked when the handler halts the JVM; `ForceCloseStorageFailureHandler` just sets a flag and doesn't drain the queue).
- Fix a test bug: `prepareObject(int count, long ttl)` was stubbed with `any()` against its primitive `int` param, which corrupted Mockito's matcher stack and broke an unrelated test (`testAppend`) in the same class. Switched to `anyInt()`.

## Test plan
- [x] `./gradlew :s3stream:test --tests "com.automq.stream.s3.S3StorageTest"` — all 4 tests pass
- [x] Verified `prepareDeltaWALUpload`'s exceptional path now drains `walPrepareQueue` and fails all queued contexts

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/S3Storage.java
@@ -773,6 +773,8 @@ public class S3Storage implements Storage {
             }
         }, backgroundExecutor).exceptionally(ex -> {
             LOGGER.error("Unexpected exception when prepare commit stream set object", ex);
+            context.cf.completeExceptionally(ex);
+            storageFailureHandler.handle(ex);
             return null;
         });
     }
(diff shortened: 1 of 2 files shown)


superhx:
@utkarshmankad ControllerObjectManager.prepareObject features unlimited retries and theoretically should never fail. In what scenario did you observe prepareObject failing?

utkarshmankad:
_prepareObject_ retries indefinitely so it shouldn't fail on its own. This catch is really for the rest of the chain: bugs in the retry logic, unexpected runtime exceptions, cancellation, etc. Right now any of those wedges _walPrepareQueue_ permanently since the context never gets polled off. 
Wasn't chasing a specific repro, just closing the hole so a failure anywhere in this path doesn't hang the pipeline.

superhx (approved):

superhx (approved):

PR #3487, opened 2026-07-31
fix(s3stream): distinguish capacity vs TTL-expiry evictions in block cache warning

### LINKED ISSUE #2662: The value of the kafka_stream_block_cache_size_bytes metric is different from the value of the s3.block.cache.size setting
We configured s3.block.cache.size=10737418240 (10GB) in server.properties, but the value of the kafka_stream_block_cache_size_bytes indicator seen in the s3 metrics exporter is different from the value set for s3.block.cache.size. The size shown by the indicator is less than 1G and it keeps changing. I would like to ask why this is the case.

Image

utkarshmankad:
## Summary
- `StreamReader` logged `"The unread block is evicted, please increase the block cache size"` for **every** unread-block eviction, regardless of cause.
- `DataBlockCache#evict0()` evicts a block for two different reasons: the cache is full (`sizeLimiter.requiredRelease()`) or the block simply sat past its TTL (`DATA_TTL`, 1 min) because the consumer is reading slower than the readahead pace. The old log message conflated both, so users chasing the warning kept bumping `s3.block.cache.size` (as reported in #2662) without effect, because the real cause was consumer lag.
- `DataBlock#free()` now takes a `DataBlock.EvictReason` (`NONE` / `CAPACITY` / `EXPIRED`). `evict0()` derives the reason from `isExpired(expiredTimestamp)` before freeing. Normal read-completion frees (`DataBlockCache#markRead`) still pass `NONE`.
- `StreamReader#handleBlockFree` now logs a distinct message for TTL-expiry evictions ("...evicted because it exceeded the cache TTL, the consumer may be reading too slowly") instead of always pointing at cache size.

Fixes #2662 (maintainer confirmed this is the right fix location: https://github.com/AutoMQ/automq/issues/2662#issuecomment-3072671443).

## Test plan
- [x] `./gradlew :s3stream:compileJava :s3stream:compileTestJava` — clean
- [x] `./gradlew :s3stream:test --tests "com.automq.stream.s3.cache.blockcache.DataBlockCacheTest" --tests "com.automq.stream.s3.cache.blockcache.StreamReaderTest"` — all pass

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/DataBlock.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/DataBlock.java
@@ -82,7 +82,7 @@ import static com.automq.stream.s3.ByteBufAlloc.BLOCK_CACHE;
      */
     public void completeExceptionally(Throwable ex) {
         loadCf.completeExceptionally(ex);
-        free0();
+        free0(EvictReason.NONE);
     }
 
     public CompletableFuture<DataBlock> dataFuture() {
@@ -90,15 +90,19 @@ import static com.automq.stream.s3.ByteBufAlloc.BLOCK_CACHE;
     }
 
     public void free() {
-        release();
-        free0();
+        free(EvictReason.NONE);
     }
 
-    private void free0() {
+    public void free(EvictReason evictReason) {
+        release();
+        free0(evictReason);
+    }
+
+    private void free0(EvictReason evictReason) {
         freeCf.complete(this);
         for (FreeListener listener : freeListeners) {
             try {
-                listener.onFree(this);
+                listener.onFree(this, evictReason);
             } catch (Throwable e) {
                 LOGGER.error("invoke onFree fail", e);
             }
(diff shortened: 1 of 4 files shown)


superhx:
There is a reachable path where the TTL eviction reason is lost, so this can still emit the misleading capacity warning from #2662.

In `DataBlock.registerFreeListener`, an already-freed block always invokes the late listener with `EvictReason.NONE`:

```java
if (freeCf.isDone()) {
    listener.onFree(this, EvictReason.NONE);
}
```

For a loaded cache hit, `DataBlockCache.Cache#getBlock0` completes `cf` and then calls `tryEvictExpired()` before returning. The latter can free that block with `EXPIRED` before `StreamReader.Block#newBlockWithData` registers its listener. The late listener then receives `NONE`, and `handleBlockFree` falls back to "please increase the block cache size."

Could we persist the actual free/eviction reason and replay it to late listeners? A regression test covering this expired-cache-hit ordering (and both resulting warning classifications) would protect the behavior requested by #2662.

utkarshmankad:
fixed. DataBlock now stashes the actual evictReason in free0(), and the already-freed branch in registerFreeListener replays that instead of hardcoding NONE — so a late listener (StreamReader, in the tryEvictExpired()-before-registration case you flagged) gets the real cause.
Added a couple tests in DataBlockTest for late registration after both EXPIRED and CAPACITY frees. Ran the block-cache test package locally

superhx (approved):

PR #3493, opened 2026-08-05
fix(autobalancer): skip metrics reporter on controller-only nodes

### LINKED ISSUE #3492: [BUG] AutoBalancerMetricsReporter fatally terminates controller-only nodes ("No listener found for listener name"), blocking controller/broker separation on Strimzi
### AutoMQ Version

`1.6.4` (`automqinc/automq:1.6.4-strimzi`). The relevant code path is unchanged on `main` (`a0a0421027`) and in `1.7.2` — see the analysis below.

### Operating System

Container image (`-strimzi` official image), running on Amazon EKS.

### Installation Method

docker — the official `-strimzi` image, deployed by the Strimzi Cluster Operator.

### Other Relevant Software

- Strimzi Cluster Operator `0.47.0`
- `Kafka.spec.kafka.version: 3.9.0`
- Kubernetes / Amazon EKS

### What Went Wrong?

A **controller-only** node (`process.roles=controller`) terminates fatally during startup as soon as `metric.reporters=kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter` is present in its config:

```
org.apache.kafka.common.config.ConfigException: No listener found for listener name: PLAIN1-9092
	at kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter.getBootstrapServers(AutoBalancerMetricsReporter.java:124)
	at kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter.configure(AutoBalancerMetricsReporter.java:231)
	...
	at kafka.server.ControllerServer.startup(ControllerServer.scala:411)
```

Note the frame: `ControllerServer.startup` → `DynamicBrokerConfig.addReconfigurables` → `DynamicMetricsReporters`. Kafka instantiates every class in `metric.reporters` on a controller-only node too, and `AutoBalancerMetricsReporter` has no notion of node role:

- `configure()` (`AutoBalancerMetricsReporter.java:230-235` on `main`) calls `getBootstrapServers(configs, autobalancer.client.listener.name)` unconditionally.
- `getBootstrapServers()` (`:101-128`) explicitly `continue`s past the `CONTROLLER` listener, then throws `ConfigException("No listener found for listener name: ...")` at `:124` when no matching non-controller listener remains.
- A controller-only node only has its control-plane listener, so the lookup can never succeed → `ProcessTerminatingFaultHandler` → process exit → `CrashLoopBackOff`.

There is no `process.roles` guard anywhere in the class, and `StaticAutoBalancerConfig` exposes only `autobalancer.client.listener.name` and `autobalancer.client.auth.sasl.jaas.config` — no enable/disable switch and no "skip when unavailable" path. I diffed `1.7.2..main` on this file: the only change is a logging tweak, so this affects the latest code as well.

#### Why this cannot be worked around on Strimzi

Issue #1746 records the current design assumption — *"the metrics reporter for autobalancer only starts in broker node"*. That holds only if the deployer can scope `metric.reporters` to broker nodes. On Strimzi, they cannot:

1. `metric.reporters` must go in `Kafka.spec.kafka.config`, which the operator renders into **every** node's config regardless of role.
2. `KafkaNodePool` has no per-pool config field. Confirmed against the CRD shipped with the operator we run (Strimzi 0.47.0, `045-Crd-kafkanodepool.yaml`, `v1beta2`): `spec` accepts only `jvmOptions, replicas, resources, roles, storage, template`.
3. Strimzi 0.47.0 does not filter `metric.reporters` out for controller-role nodes — the stack trace above comes from a node the operator rendered as controller-only.

So the only configuration that boots is combined `roles: [controller, broker]` node pools, which is what we run in production today. That means we cannot separate the control plane from the data plane at all: every controller must also carry broker traffic, and every broker-pool scaling decision is constrained by the controller quorum. Given that AutoMQ publishes an official `-strimzi` image (`.github/workflows/docker-strimzi-release.yaml`) and Strimzi is an accepted deployment path (#2287), this looks like a gap on the AutoMQ side rather than an operator problem.

Independently of Strimzi: a config value that is legal cluster-wide causing a `ProcessTerminatingFaultHandler` kill on a controller seems like a sharp edge worth removing regardless of how the config got there.

### What Should Have Happened Instead?

`AutoBalancerMetricsReporter` should be a no-op on a node without the broker role, rather than killing the process. Either:

1. **Role guard in `configure()`** — read `process.roles` (the same config map already provides `node.id`, read at `:260`); if the node has no `broker` role, log at INFO and skip producer setup / do not start the reporter thread. This matches the design intent stated in #1746 and makes it true by construction instead of by deployment convention.
2. **Degrade instead of throw** — if `getBootstrapServers()` finds no non-`CONTROLLER` listener matching `autobalancer.client.listener.name`, log a warning and disable the reporter instead of throwing `ConfigException`.

Option 1 seems more precise; option 2 is the smaller change.

### Steps to Reproduce

1. Deploy the Strimzi Cluster Operator `0.47.0` with the AutoMQ `-strimzi` image mapped for Kafka `3.9.0`.
2. Create a `Kafka` CR (redacted to the relevant keys):

```yaml
spec:
  kafka:
    version: 3.9.0
    listeners:
      - name: plain1
        port: 9092
        type: internal
        tls: false
    config:
      autobalancer.controller.enable: "true"
      autobalancer.client.listener.name: PLAIN1-9092
      metric.reporters: kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter
```

3. Create two `KafkaNodePool`s for that cluster:
   - pool A: `roles: [controller]`
   - pool B: `roles: [broker]`
4. Pool B (brokers) starts normally. Pool A (controllers) crash-loops with the stack trace below.
5. Change pool A to `roles: [controller, broker]` → the cluster starts, because the node now also has the `PLAIN1-9092` listener. This combined-role layout is the only configuration we have found that works.

### Additional Information

Full stack trace from a controller-only pod:

```
2026-07-03 10:30:27,7xx ... (kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter) [main]
2026-07-03 10:30:27,723 ERROR Encountered fatal fault: caught exception (org.apache.kafka.server.fault.ProcessTerminatingFaultHandler) [main]
org.apache.kafka.common.config.ConfigException: No listener found for listener name: PLAIN1-9092
	at kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter.getBootstrapServers(AutoBalancerMetricsReporter.java:124)
	at kafka.autobalancer.metricsreporter.AutoBalancerMetricsReporter.configure(AutoBalancerMetricsReporter.java:231)
	at org.apache.kafka.common.config.AbstractConfig.getConfiguredInstance(AbstractConfig.java:403)
	at org.apache.kafka.common.config.AbstractConfig.getConfiguredInstances(AbstractConfig.java:484)
	at kafka.server.DynamicMetricReporterState.createReporters(DynamicBrokerConfig.scala:937)
	at kafka.server.DynamicMetricReporterState.<init>(DynamicBrokerConfig.scala:923)
	at kafka.server.DynamicMetricsReporters.<init>(DynamicBrokerConfig.scala:863)
	at kafka.server.DynamicBrokerConfig.addReconfigurables(DynamicBrokerConfig.scala:305)
	at kafka.server.ControllerServer.startup(ControllerServer.scala:411)
	at kafka.server.KafkaRaftServer.$anonfun$startup$1(KafkaRaftServer.scala:87)
	at kafka.server.KafkaRaftServer.$anonfun$startup$1$adapted(KafkaRaftServer.scala:87)
	at scala.Option.foreach(Option.scala:437)
	at kafka.server.KafkaRaftServer.startup(KafkaRaftServer.scala:87)
	at kafka.Kafka$.main(Kafka.scala:160)
	at kafka.Kafka.main(Kafka.scala)
```

Related: #1746 (states the broker-only design assumption), #2287 (Strimzi as a deployment path).

**Questions**

1. Is controller/broker role separation intended to be supported when AutoBalancer is enabled? Or is combined `[controller, broker]` the expected topology for AutoMQ + Strimzi?
2. If separation is intended, would you accept a PR adding the `process.roles` guard in `AutoBalancerMetricsReporter#configure()`? Happy to open one.

BackendArchitectX:
## Summary

Fixes AutoBalancerMetricsReporter startup on controller-only nodes.

- Skips reporter configuration when the broker role is absent
- Avoids broker listener resolution and producer creation on controller-only nodes
- Prevents the reporter thread from starting on controller-only nodes
- Preserves existing behavior for broker, combined-role, and missing-role configurations
- Adds regression coverage for all role scenarios

Closes #3492

## Testing

-  `AutoBalancerMetricsReporterTest` — passed
- `:core:checkstyleMain :core:checkstyleTest` — passed
- `:core:spotlessJavaCheck` — passed
- `jar -x test` — passed
- `core:S3UnitTest` — all AutoBalancerMetricsReporter tests passed; 4 unrelated Windows filesystem tests failed due to file-lock/path behavior

### DIFF
--- a/core/src/main/java/kafka/autobalancer/metricsreporter/AutoBalancerMetricsReporter.java
+++ b/core/src/main/java/kafka/autobalancer/metricsreporter/AutoBalancerMetricsReporter.java
@@ -96,6 +96,46 @@ public class AutoBalancerMetricsReporter implements MetricsRegistryListener, Met
     private volatile boolean shutdown = false;
     private int metricsReporterCreateRetries;
     private long lastErrorReportTime = 0;
+    private boolean enabled = true;
+
+    private boolean isControllerOnly(Object processRoles) {
+        if (processRoles == null) {
+            return false;
+        }
+
+        if (processRoles instanceof Iterable<?>) {
+            boolean hasRole = false;
+            for (Object role : (Iterable<?>) processRoles) {
+                String roleName = String.valueOf(role).trim();
+                if (roleName.isEmpty()) {
+                    continue;
+                }
+                hasRole = true;
+                if ("broker".equalsIgnoreCase(roleName)) {
+                    return false;
+                }
+            }
+            return hasRole;
+        }
+
+        String roles = String.valueOf(processRoles).trim();
+        if (roles.isEmpty()) {
+            return false;
+        }
+
+        boolean hasRole = false;
+        for (String role : roles.split(",")) {
+            String roleName = role.trim();
+            if (roleName.isEmpty()) {
+                continue;
+            }
+            hasRole = true;
+            if ("broker".equalsIgnoreCase(roleName)) {
+                return false;
+            }
+        }
+        return hasRole;
+    }
 
     String getBootstrapServers(Map<String, ?> configs, String expectedListenerName) {
         String listenerStr = String.valueOf(configs.get(SocketServerConfigs.LISTENERS_CONFIG));
@@ -128,6 +168,9 @@ public class AutoBalancerMetricsReporter implements MetricsRegistryListener, Met
 
     @Override
     public void init(List<KafkaMetric> metrics) {
+        if (!enabled) {
+            return;
+        }
         metricsReporterRunner = new KafkaThread("AutoBalancerMetricsReporterRunner", this, true);
         yammerMetricProcessor = new YammerMetricProcessor();
         metricsReporterRunner.start();
@@ -221,6 +264,12 @@ public class AutoBalancerMetricsReporter implements MetricsRegistryListener, Met
     public void configure(Map<String, ?> rawConfigs) {
 
         Map<String, Object> configs = new HashMap<>(rawConfigs);
+
+        if (isControllerOnly(configs.get(KRaftConfigs.PROCESS_ROLES_CONFIG))) {
+            enabled = false;
+            LOGGER.info("Skipping AutoBalancerMetricsReporter on controller-only node");
+            return;
+        }
 
         StaticAutoBalancerConfig staticAutoBalancerConfig = new StaticAutoBalancerConfig(configs, false);
         Properties producerProps = AutoBalancerMetricsReporterConfig.parseProducerConfigs(configs);
--- a/core/src/main/java/kafka/autobalancer/metricsreporter/AutoBalancerMetricsReporter.java
+++ b/core/src/main/java/kafka/autobalancer/metricsreporter/AutoBalancerMetricsReporter.java
@@ -39,6 +39,7 @@ import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.common.errors.InterruptException;
 import org.apache.kafka.common.internals.Topic;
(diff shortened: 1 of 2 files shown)


superhx:
I think this logic and its tests can be simplified.

`isControllerOnly` does not depend on any reporter instance state, so it should be a package-private `static` method. Kafka creates metric reporters with `KafkaConfig.originals()`: normal `server.properties`/Strimzi startup therefore supplies `process.roles` as a raw string such as `"controller"` or `"broker,controller"`, while programmatic configuration may supply a `List`. Kafka already defines this property as `ConfigDef.Type.LIST`, whose supported input forms are exactly `String` and `List`, so we can reuse that parser instead of handling arbitrary `Iterable` values and arbitrary object stringification:

```java
static boolean isControllerOnly(Object processRoles) {
    if (processRoles == null) {
        return false;
    }
    List<?> roles = (List<?>) ConfigDef.parseType(
        KRaftConfigs.PROCESS_ROLES_CONFIG,
        processRoles,
        ConfigDef.Type.LIST);
    return roles.equals(List.of("controller"));
}
```

The role-input matrix can then be covered directly with parameterized tests for this static helper, without constructing a reporter or providing unrelated listener/node configuration. In particular, the combined-role production case should use the raw string `"broker,controller"` rather than only `List.of("broker", "controller")`.

I would still keep one higher-level controller-only regression test around `configure()` + `init()`. That test protects the actual bug fix: configuration must return before listener resolution/producer creation, and `init()` must not start the reporter thread. The remaining broker/combined/missing-role cases can be focused helper tests.

BackendArchitectX:
Updated, thanks for the suggestion. `isControllerOnly` is now package-private static and uses `ConfigDef.parseType` with `ConfigDef.Type.LIST`. I replaced the role-specific reporter tests with parameterized coverage for String/List inputs, including the raw `broker,controller` case, while keeping the controller-only `configure()` + `init()` regression test.

The focused test suite passes on Java 17, along with core Checkstyle, Spotless, and the jar build.

superhx (approved):

PR #3495, opened 2026-08-07
feat(stream): support fast partition reassignment

superhx:
## Why

For a healthy RF=1 partition reassignment, the source broker currently follows
the normal stream-close path:

1. Stop new writes and drain in-flight appends.
2. Persist the final partition and MetaStream metadata.
3. Upload outstanding WAL data to ObjectStorage and commit the corresponding
   object metadata.
4. Close the streams and allow the target broker to reopen them.
5. Replay MetaStream records on the target to recover the partition.

The source cannot complete stream close until the outstanding WAL upload and
object metadata commit finish. In observed environments, this part of the close
path can take close to one second.

The target also needs the latest MetaStream state before it can reconstruct the
partition. Recovering that state through ObjectStorage puts remote PUT, metadata
commit, and GET latency directly on the partition write-unavailability path.

This makes reassignment latency unnecessarily dependent on ObjectStorage even
when both the source and target brokers are healthy.

## What

This PR introduces fast partition reassignment for AutoMQVersion V6.

The source and target still need to exchange the final MetaStream state required
to reconstruct the partition, including:

- log segment metadata;
- partition metadata;
- leader epoch checkpoints;
- producer snapshot metadata;
- unknown MetaStream keys required for forward compatibility;
- the exact MetaStream handoff end offset.

Instead of waiting for this state to become available through ObjectStorage, the
source freezes an immutable latest-KV MetaStream snapshot and sends it directly
to the target broker through a partition handoff RPC.

The target can use this handoff to recover the MetaStream state after Controller
open authorization, removing synchronous WAL upload/commit and MetaStream
ObjectStorage replay from the write-readiness critical path.

Outstanding source WAL data is still uploaded and committed asynchronously.
The optimization changes when the target may resume ownership; it does not
reduce ObjectStorage durability or read coverage.

### 1. Partition handoff lifecycle

The source broker keeps the existing graceful-close prefix:

- reject new partition writes;
- drain in-flight appends;
- persist the producer snapshot and final partition metadata;
- flush final segment and index metadata.

After the final MetaStream append completes, the source freezes the MetaStream.
A frozen MetaStream is immutable and provides an internally consistent
latest-value snapshot together with its exclusive end offset.

The source then:

1. Prepares the handoff on the target through the new
   `AUTOMQ_PREPARE_PARTITION_HANDOFF` broker RPC.
2. Initiates data-stream close and MetaStream close through the existing ordered
   Controller request path.
3. Fast-closes each stream at its logical end offset.
4. Starts the remaining WAL upload in the background without waiting for it on
   the reassignment path.

Prepared handoffs are held in a bounded, TTL-based, get-once cache on the target.

After the Controller authorizes the target to open the MetaStream, the target
consumes the exact handoff identified by topic ID, partition ID, and MetaStream
end offset. The handoff is replayed into temporary state and published
atomically. The target then opens the data streams and continues the normal
partition initialization flow.

The handoff is only a performance hint. A missing, expired, evicted, oversized,
or failed handoff falls back to the existing ObjectStorage replay path.

### 2. Metadata commit and failover adaptation

Fast close separates stream ownership progress from ObjectStorage durability:

- `stream.endOffset` becomes the logical ownership and append boundary.
- Object metadata remains the authority for actual ObjectStorage read coverage.
- A new node-scoped WAL-uncommitted metadata entry records the sealed historical
  range that the former owner still has to upload or recover.

When a stream is fast-closed, the Controller atomically:

- advances the stream logical end;
- seals the source ownership range;
- records the source node's remaining historical WAL responsibility.

Subsequent object commits are classified by ownership:

- a historical owner may only advance its own sealed uncommitted range;
- the current owner advances the stream logical end through normal commits;
- gaps, visible overlaps, stale owners, and commits beyond a sealed boundary are
  rejected.

Trim evaluates historical responsibility lazily against the current stream
start offset. It does not weaken object coverage or scan every node entry.

For node failover, `getOpeningStreams` returns both currently opened streams and
active historical WAL ranges owned by the failed node. Existing WAL recovery can
therefore upload and commit the remaining historical data independently of the
current partition owner.

An active historical entry also prevents the same node from reopening the stream
as a new owner until its previous responsibility is resolved. This protects
A-to-B-to-A reassignment from giving one node multiple disjoint recoverable WAL
ranges for the same stream.

## Correctness and compatibility

- Handoff does not grant stream-open permission; Controller metadata remains the
  ownership authority.
- Handoff is not a durability signal.
- Reads still require continuous LogCache, BlockCache, or ObjectStorage metadata
  coverage and never skip an uncommitted historical range.
- Source WAL upload continues asynchronously and is drained during clean broker
  shutdown.
- Cache miss, broker restart, RPC failure, or replay failure falls back to the
  existing ObjectStorage recovery path.
- V5 keeps the legacy close and reassignment behavior.
- Fast reassignment and its metadata records are enabled only after
  AutoMQVersion V6 is finalized.
- Direct downgrade from V6 to V5 is rejected because V5 brokers cannot interpret
  the new lifecycle metadata safely.

## Review guide

The implementation is best reviewed in this order:

1. MetaStream freeze, handoff RPC, target cache, and atomic replay.
2. Fast stream close and logical-end semantics.
3. Historical WAL responsibility metadata and object commit classification.
4. Trim, reopen, and node-failover handling.
5. AutoMQVersion gating and downgrade protection.
6. Partition lifecycle scheduling and supporting latency optimizations.
7. Focused Controller, metadata image, stream, and lifecycle tests.


1sonofqiu (approved):

daniel-y (commented):
Reviewed the partition-migration consistency, close/open ordering, bulk-operation, failure-recovery, and resource-lifecycle paths. I found three issues that can block reassignment or make it livelock; details are inline.

core/src/main/scala/kafka/log/streamaspect/ElasticLog.scala
```diff
@@ -500,31 +509,55 @@ class ElasticLog(val metaStream: MetaStream,
...
 
-    /**
-     * Directly close all streams of the log.
-     */
-    def closeStreams(): CompletableFuture[Void] = {
+    private[streamaspect] def closeStreams(fastClose: Boolean): CompletableFuture[Void] = {
+        // Initiate every data-stream close before starting the MetaStream handoff and close. The normal RPC batching
+        // and Controller processing path therefore closes data-stream ownership before the later MetaStream close
+        // completes. The target starts replay and data-stream open only after observing that MetaStream close.
+        val dataStreamsCloseFuture = streamManager.close()
```
daniel-y, line 555: **[P1] The MetaStream close is not actually ordered after data-stream closes.** `streamManager.close()` only initiates asynchronous close on every stream; each `S3Stream.close()` still waits for pending append/fetch/trim before sending its Controller close. Meanwhile, `metaStream.close(fastClose)` may complete the handoff and reach CLOSED in the Controller first. As soon as the destination observes the MetaStream as CLOSED, it starts opening data streams, and any stream still owned by the source keeps returning `STREAM_NOT_CLOSED`. Please chain the MetaStream close after `dataStreamsCloseFuture` completes successfully, for example with `thenCompose`. Fast close already skips waiting for object upload, so this preserves the main speedup.

core/src/main/scala/kafka/log/streamaspect/ElasticLog.scala
```diff
@@ -500,31 +509,55 @@ class ElasticLog(val metaStream: MetaStream,
...
+                    warn(s"${logIdent}failed to persist final metadata; closing without handoff", e)
+                    CompletableFuture.failedFuture(e)
+            }
+            flushFuture.handle[Boolean]((_, exception) => {
+                if (exception != null) {
+                    warn(s"${logIdent}failed to flush the active segment; closing without handoff",
+                        FutureUtil.cause(exception))
+                }
+                exception == null
+            }).thenCompose(fastClose => closeStreams(fastClose)).handle[Void]((_, exception) => {
```
daniel-y, line 540: **[P1] Swallowing a terminal stream-close failure here can leave the partition permanently stuck in reassignment.** The source partition has already been removed from `allPartitions` before close is invoked. If the Controller close for any data stream ultimately fails, this handler still returns successfully and no component retries that close. The destination then retries `STREAM_NOT_CLOSED` forever for the stream that remains OPENED, usually until the source broker restarts and performs WAL recovery. Please propagate terminal failures and provide an explicit close-retry or broker fail-fast path; sending the handoff is not sufficient to consider ownership safely transferred.

core/src/main/scala/kafka/log/stream/s3/network/ControllerRequestSender.java
```diff
@@ -175,11 +178,16 @@ synchronized void send(RequestTask task) {
...
+                if (lingerNanos == 0) {
+                    send0();
+                } else {
+                    retryService.schedule(this::send0, lingerNanos, TimeUnit.NANOSECONDS);
+                }
             }
         }
 
-        void send0() {
+        synchronized void send0() {
```
daniel-y, line 190: **[P1] With the new linger, this still drains the entire queue without a bound.** Open/close requests are aggregated into one atomic Controller result, while the Controller caps a metadata batch at 50,000 records. A reopen normally emits three records, so roughly 16.7k streams can deterministically exceed the limit. The entire batch then receives a generic error; the client retries indefinitely and is likely to combine the same tasks into another oversized batch, causing migration livelock. Please split batches by subrequest count, encoded byte size, and expected metadata-record count, and support split/binary retry for batch-too-large or generic failures.

core/src/main/scala/kafka/log/streamaspect/ElasticLog.scala
superhx: In most cases, the data stream is closed before the meta stream, and latency‑optimization is performed based on this assumption. Even if the meta stream is closed before the data stream, there will be no correctness issues.

core/src/main/scala/kafka/log/stream/s3/network/ControllerRequestSender.java
superhx: This is a great find. The next PR will uniformly enhance ControllerRequestSender to provide a limit for the batching upper bound.

PR #3505, opened 2026-08-11
feat(controller): support gentle elastic broker shutdown

superhx:
## Why

An AutoMQ Broker can lead a large number of Partitions. Migrating all leadership in one controlled-shutdown heartbeat produces a large metadata burst. Controller changes must also preserve completed progress and continue draining the remaining leadership.

## What

Introduce gentle controlled shutdown that migrates leadership in periodic batches. Multiple Brokers drain concurrently with independent batch budgets, and a newly active Controller reconstructs ephemeral pacing state from the remaining Leader Partitions.

Upstream Kafka controlled-shutdown behavior remains unchanged outside AutoMQ mode.

## How

A dedicated `GentleControlledShutdownControlManager` uses the shared scheduler to submit a Controller write event once per second while the Controller is active. The write event scans persisted controlled-shutdown Broker registrations and drains one independent batch for each Broker.

Each Broker batch target is computed once per Controller lifecycle as the larger of 100 Partitions or approximately one sixtieth of its remaining leaders. Actual Partition changes remain owned by `ReplicationControlManager`, which keeps same-topic work contiguous where possible and advances the controlled-shutdown offset barrier from the committed batch-end offset.

Controller activation and renounce clear ephemeral per-Broker pacing state. Scheduling is non-blocking; inactive Controllers skip submission, and shutdown cancels the manager task without closing the shared scheduler.

## Reviewer notes

Suggested reading order:

1. `GentleControlledShutdownControlManager` for scheduling, lifecycle, concurrent Broker pacing, and barrier advancement.
2. `ReplicationControlManager` for controlled-shutdown deferral and bounded Partition migration.
3. `BrokerHeartbeatManager` for restoring and advancing the ephemeral controlled-shutdown offset barrier.
4. The AutoMQ-specific `S3Unit` tests for fixed and scaled batches, topic grouping, concurrent drains, lifecycle recomputation, and Kafka-compatible fallback behavior.

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/BrokerHeartbeatManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/BrokerHeartbeatManager.java
@@ -471,6 +471,29 @@ public class BrokerHeartbeatManager {
         }
     }
 
+    // AutoMQ inject start
+    /**
+     * Establish or advance the metadata barrier for a Broker in persisted controlled shutdown.
+     * Establishing the barrier restores heartbeat soft state after a Controller change.
+     */
+    void advanceControlledShutdownOffset(int brokerId, long controlledShutdownOffset) {
+        BrokerHeartbeatState broker = heartbeatStateOrThrow(brokerId);
+        if (broker.fenced()) {
+            throw new IllegalStateException("Fenced broker " + brokerId
+                + " cannot advance its controlled shutdown offset.");
+        }
+        active.remove(broker);
+        if (!broker.shuttingDown()) {
+            broker.lastControlledShutdownNs = time.nanoseconds();
+        }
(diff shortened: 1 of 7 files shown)


superhx:
Local verification:

- `./gradlew --build-cache metadata:S3UnitTest core:S3UnitTest s3stream:test` — passed
- JDK 17: `./gradlew --build-cache rat checkstyleMain checkstyleTest spotlessJavaCheck` — relevant checks passed (`rat` is project-disabled and skipped)
- After the CI import-order finding, JDK 17: `./gradlew --build-cache metadata:spotlessJavaCheck metadata:checkstyleMain metadata:checkstyleTest` — passed

Gezi-lzq (approved):

PR #3509, opened 2026-08-12
feat(controller): support gentle controlled shutdown

superhx:
## Why

Backport the gentle controlled-shutdown behavior merged into `1.8` by #3505 to `main`.

A Broker can lead a large number of Partitions. Migrating all leadership in one controlled-shutdown heartbeat produces a large metadata burst. Controller changes must preserve completed progress and continue draining the remaining leadership.

## What

Migrate controlled-shutdown leadership in periodic batches. Multiple Brokers drain concurrently with independent batch budgets, and a newly active Controller reconstructs ephemeral pacing state from the remaining Leader Partitions.

Upstream Kafka controlled-shutdown behavior remains unchanged outside AutoMQ mode.

## How

`GentleControlledShutdownControlManager` submits a non-blocking Controller write event once per second while the Controller is active. Each Broker receives an independent batch sized to complete in approximately 60 dispatches, with a minimum of 100 Partitions per batch.

`ReplicationControlManager` owns the bounded Partition changes and keeps same-topic work contiguous where possible. `BrokerHeartbeatManager` advances the controlled-shutdown offset barrier from the committed batch-end offset.

This cherry-pick required one main-only test adaptation: `KafkaConfigSchema.EMPTY` replaces the `FakeKafkaConfigSchema` fixture that is no longer present on main.

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/BrokerHeartbeatManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/BrokerHeartbeatManager.java
@@ -471,6 +471,29 @@ public class BrokerHeartbeatManager {
         }
     }
 
+    // AutoMQ inject start
+    /**
+     * Establish or advance the metadata barrier for a Broker in persisted controlled shutdown.
+     * Establishing the barrier restores heartbeat soft state after a Controller change.
+     */
+    void advanceControlledShutdownOffset(int brokerId, long controlledShutdownOffset) {
+        BrokerHeartbeatState broker = heartbeatStateOrThrow(brokerId);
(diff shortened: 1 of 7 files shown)


superhx:
Local verification on JDK 17:

- `./gradlew --build-cache metadata:S3UnitTest core:S3UnitTest s3stream:test` — passed
- `./gradlew --build-cache metadata:spotlessJavaCheck metadata:checkstyleMain metadata:checkstyleTest` — passed

Gezi-lzq (approved):

Gezi-lzq (approved):

PR #3527, opened 2026-08-16
fix(s3stream): avoid multipart upload overhead for small MINOR_V1 copyWrite objects

ashsamantaray-stripe:
## Summary
- `ProxyWriter.copyWrite()` unconditionally escalated to `MultiPartWriter` on the very first call, regardless of the copied range's size. For MINOR_V1 stream object compaction, source objects are always below `Writer.MIN_PART_SIZE` (5MiB), so every compaction group paid the full multipart upload overhead (`CreateMultipartUpload` + `CompleteMultipartUpload`, 2 extra requests) instead of a single `PutObject`, even though the merged group total is small enough to fit comfortably under `ProxyWriter`'s existing single-PUT threshold (`MAX_UPLOAD_SIZE`, 32MiB).
- This change makes `ProxyWriter.copyWrite()` behave consistently with `write()`: ranges below `minPartSize` (which can't use a real server-side `UploadPartCopy` anyway) are read via `rangeRead` and buffered into the same single-PUT `ObjectWriter` path, only escalating to `MultiPartWriter` once the accumulated size crosses `MAX_UPLOAD_SIZE`. Ranges at/above `minPartSize` keep using the existing zero-download `UploadPartCopy` fast path unchanged.

## Test plan
- [x] `:s3stream:test --tests ProxyWriterTest` — added `testCopyWrite_smallObjectsUseSinglePut` verifying multiple small copyWrite calls result in exactly one `PutObject` and zero multipart API calls.
- [x] `:s3stream:test --tests MultiPartWriterTest` — unaffected, still green.
- [x] `:s3stream:test --tests com.automq.stream.s3.compact.*` — unaffected, still green.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/operator/ProxyWriter.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/operator/ProxyWriter.java
@@ -79,10 +79,23 @@ public class ProxyWriter implements Writer {
 
     @Override
     public void copyWrite(S3ObjectMetadata s3ObjectMetadata, long start, long end) {
-        if (largeObjectWriter == null) {
+        if (largeObjectWriter != null) {
+            largeObjectWriter.copyWrite(s3ObjectMetadata, start, end);
+            return;
+        }
+        if (end - start >= minPartSize) {
+            // Large enough to benefit from a zero-download server-side UploadPartCopy, so it's worth paying
+            // the multipart upload overhead (createMultipartUpload/completeMultipartUpload).
+            newLargeObjectWriter(writeOptions, objectStorage, path);
+            largeObjectWriter.copyWrite(s3ObjectMetadata, start, end);
+            return;
+        }
+        // Below minPartSize, S3 can't do a server-side UploadPartCopy for it anyway, so just buffer the range
+        // read like write() does and let it ride the single PutObject path instead of forcing multipart.
+        objectWriter.copyWrite(s3ObjectMetadata, start, end);
+        if (objectWriter.isFull()) {
             newLargeObjectWriter(writeOptions, objectStorage, path);
         }
-        largeObjectWriter.copyWrite(s3ObjectMetadata, start, end);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


superhx (dismissed):

s3stream/src/main/java/com/automq/stream/s3/operator/ProxyWriter.java
```diff
@@ -119,12 +132,17 @@ public short bucketId() {
 
     protected void newLargeObjectWriter(WriteOptions writeOptions, ObjectStorage objectStorage, String path) {
         this.largeObjectWriter = new MultiPartWriter(writeOptions, objectStorage, path, minPartSize);
-        if (objectWriter.data.readableBytes() > 0) {
-            FutureUtil.propagate(largeObjectWriter.write(objectWriter.data), objectWriter.cf);
-        } else {
-            objectWriter.data.release();
-            objectWriter.cf.complete(null);
-        }
+        // wait for any in-flight copyWrite range reads to land in objectWriter.data before handing it off.
```
superhx: The waiting here is asynchronous, meaning data written later by ProxyWriter may appear ahead of that from MultiPartWriter. Perhaps we can turn copyWrite and write of ObjectWrite into waiting operations, which are either executed upon close, or replayed to MultiPartWriter by newLargeObjectWriter.

s3stream/src/main/java/com/automq/stream/s3/operator/ProxyWriter.java
```diff
@@ -119,12 +164,25 @@ public short bucketId() {
...
-        }
+        // Hand off everything already buffered in objectWriter as the first multipart operation, then serialize
+        // every subsequent largeObjectWriter call onto this chain (see copyWriteToLargeObject/writeToLargeObject).
+        // This guarantees two things:
+        //   (a) parts are issued in the exact order copyWrite/write were called. Byte order in the final object
+        //       must match, since the compaction index block encodes absolute byte offsets into the object.
+        //   (b) MultiPartWriter, which assumes a single-threaded caller, is only ever touched by one thread at a
+        //       time - even though the handoff below waits on range reads that complete on a different thread.
+        // whenComplete keeps mpwChain mirroring lastOpCf, so a failed range read short-circuits the whole chain.
+        this.mpwChain = objectWriter.lastOpCf.whenComplete((nil, ex) -> {
```
ashsamantaray-stripe, line 175: While the handoff will not happen for MINOR_V1, we preserve in the case where it does happen for MINOR/MAJOR. The handoff waits on all tasks attached in lastOpCf to complete and then passes that data to largeObjectWriter if possible.

s3stream/src/main/java/com/automq/stream/s3/operator/ProxyWriter.java
```diff
@@ -79,10 +82,49 @@ public void copyOnWrite() {
...
+            // Large enough to benefit from a zero-download server-side UploadPartCopy, so it's worth paying
+            // the multipart upload overhead (createMultipartUpload/completeMultipartUpload).
+            newLargeObjectWriter(writeOptions, objectStorage, path);
+            copyWriteToLargeObject(s3ObjectMetadata, start, end);
+            return;
+        }
+        // Below minPartSize, S3 can't do a server-side UploadPartCopy for it anyway, so just buffer the range
+        // read like write() does and let it ride the single PutObject path instead of forcing multipart.
+        objectWriter.copyWrite(s3ObjectMetadata, start, end);
+        if (objectWriter.isFull()) {
```
ashsamantaray-stripe, line 99: ObjectWriter's buffered data exceeds 32MiB then we switch to LargeObjectWriter, triggering hand-off in the constructor itself

s3stream/src/main/java/com/automq/stream/s3/operator/ProxyWriter.java
ashsamantaray-stripe: Updated: Using chained callbacks to synchronize read-write cycles. We handle the cleanup on close()

superhx (approved):

PR #3539, opened 2026-08-21
fix(metadata): schedule prepared object timeout at deadline

### LINKED ISSUE #3523: [BUG] Prepared S3 object TTL checks are delayed by an incorrect timeout calculation
### AutoMQ Version

main at commit 92eee56d308f72f1265820da2d3ae2065cca3796 (3.9.0-SNAPSHOT)

### Operating System

Windows 11 for source-level verification. The affected controller scheduling logic is OS-independent.

### Installation Method

source

### Hardware Configuration

N/A — this is a source-level timeout scheduling issue and is not hardware-dependent.

### Other Relevant Software

- Kafka version: 3.9.0
- Gradle version: 8.8
- Java version used for local verification: 23
- Timer implementation: Netty HashedWheelTimer

### What Went Wrong?

`S3ObjectControlManager` stores the expiration timestamp of a prepared object as an absolute deadline:

```java
long expiredTs = now + request.timeToLiveInMs();
However, newPreparedObjectTimeout calculates the timer delay using:
Math.max(
    time.milliseconds() - deadline,
    TimeUnit.MINUTES.toMillis(30)
)
```
The subtraction appears to be reversed.
When the deadline is in the future, time.milliseconds() - deadline is negative. As a result, the timeout is always scheduled with a delay of at least 30 minutes.
This means:
A prepared object with a TTL shorter than 30 minutes is not checked near its requested expiration time.
An already expired object restored during metadata replay may be delayed for another 30 minutes or longer.
Uncommitted objects may remain in the PREPARED state longer than requested.
Metadata and physical object cleanup may be delayed, increasing object-storage usage and cost.
The existing expiration test manually invokes handlePreparedObjectTimeout, so it bypasses the real timer scheduling behavior and does not detect this problem.

### What Should Have Happened Instead?

The timeout should be scheduled based on the remaining time until the absolute deadline.

Conceptually:

```java
Math.max(deadline - time.milliseconds(), 0L)
```
A prepared object should become eligible for lifecycle cleanup near the TTL requested through PrepareS3ObjectRequestData.timeToLiveInMs().
If the 30-minute minimum delay is intentional, the minimum supported TTL should instead be explicitly validated and documented so that callers cannot request a TTL that the controller does not honor.

### Steps to Reproduce

1. Check out AutoMQ commit `92eee56d308f72f1265820da2d3ae2065cca3796`.
2. Create an `S3ObjectControlManager` with `MockTime`.
3. Prepare and replay one object with a short TTL, for example 1 second.
4. Advance `MockTime` beyond the object's deadline.
5. Do not manually invoke `handlePreparedObjectTimeout`.
6. Wait a few seconds for the real `HashedWheelTimer` task.
7. Observe that the object is not added to `waitingDeadlineCheckPreparedObjects` and no expiration record is generated.

At replay time, a deadline one second in the future results in:

```text
Math.max(now - deadline, 30 minutes) = 30 minutes
```
Therefore, the timer is scheduled for 30 minutes instead of approximately one second.
A regression test that waits up to five seconds for automatic timeout delivery fails with the current implementation.

### Additional Information

```markdown
Relevant source locations:

- Deadline creation:
  https://github.com/AutoMQ/automq/blob/92eee56d308f72f1265820da2d3ae2065cca3796/metadata/src/main/java/org/apache/kafka/controller/stream/S3ObjectControlManager.java#L226-L240

- Timeout scheduling:
  https://github.com/AutoMQ/automq/blob/92eee56d308f72f1265820da2d3ae2065cca3796/metadata/src/main/java/org/apache/kafka/controller/stream/S3ObjectControlManager.java#L546-L550

- Existing expiration test:
  https://github.com/AutoMQ/automq/blob/92eee56d308f72f1265820da2d3ae2065cca3796/metadata/src/test/java/org/apache/kafka/controller/stream/S3ObjectControlManagerTest.java#L211-L244

Before preparing a PR, I would like to confirm whether the 30-minute lower bound is an intentional lifecycle policy.

If it is not intentional, I can submit a focused PR that:

- calculates the delay from `deadline - currentTime`;
- preserves the existing timer and controller-event architecture;
- adds an `S3Unit` regression test exercising the actual timer;
- makes no metadata-format, API, or configuration changes.

### PR DESCRIPTION
**Pull request by hachiwar**
## Why

Prepared S3 objects store their TTL expiration as an absolute deadline. The timer delay currently subtracts that deadline from the current time and applies a 30-minute lower bound, so future and already-expired objects are not checked at the intended time.

The existing expiration test invokes the timeout handler directly and therefore does not cover timer scheduling.

## What

Schedule each prepared-object timeout from the remaining time until its absolute deadline:

```java
Math.max(deadline - time.milliseconds(), 0L)
```

Short TTLs are delivered near their deadline, while already-expired objects are scheduled immediately. The metadata format, public API, and lifecycle event architecture remain unchanged.

## How

The existing per-object timer still owns delivery, and its callback still only enqueues the object ID for controller-event processing. A regression test prepares an object with a short TTL and waits for the real timer callback to enqueue it without invoking `handlePreparedObjectTimeout` manually.

## Reviewer notes

Review `S3ObjectControlManager.newPreparedObjectTimeout` for the delay calculation and `S3ObjectControlManagerTest.testPreparedObjectTimeoutUsesDeadline` for the timer-path regression coverage.

Fixes #3523

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/stream/S3ObjectControlManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/stream/S3ObjectControlManager.java
@@ -545,7 +545,7 @@ public class S3ObjectControlManager {
 
     private Timeout newPreparedObjectTimeout(long objectId, long deadline) {
         return preparedObjectsTimer.newTimeout(
-            t -> handlePreparedObjectTimeout(objectId), Math.max(time.milliseconds() - deadline, TimeUnit.MINUTES.toMillis(30)),
+            t -> handlePreparedObjectTimeout(objectId), Math.max(deadline - time.milliseconds(), 0L),
             TimeUnit.MILLISECONDS
         );
     }
--- a/metadata/src/test/java/org/apache/kafka/controller/stream/S3ObjectControlManagerTest.java
+++ b/metadata/src/test/java/org/apache/kafka/controller/stream/S3ObjectControlManagerTest.java
@@ -37,6 +37,7 @@ import org.apache.kafka.metadata.stream.S3Object;
 import org.apache.kafka.metadata.stream.S3ObjectState;
 import org.apache.kafka.server.common.ApiMessageAndVersion;
 import org.apache.kafka.server.common.automq.AutoMQVersion;
+import org.apache.kafka.test.TestUtils;
 import org.apache.kafka.timeline.SnapshotRegistry;
 
 import com.automq.stream.s3.Config;
@@ -139,6 +140,20 @@ public class S3ObjectControlManagerTest {
             assertEquals(time.milliseconds() + 60 * 1000, s3Object.getTimestamp());
         });
         assertEquals(8, manager.nextAssignedObjectId());
+    }
+
+    /**
+     * Given a prepared object with a short TTL, the timeout should be delivered near its deadline.
+     */
+    @Test
+    public void testPreparedObjectTimeoutUsesDeadline() throws InterruptedException {
+        long objectId = prepareOneObject(100L);
+
+        TestUtils.waitForCondition(
+            () -> manager.waitingDeadlineCheckPreparedObjects.contains(objectId),
+            5_000L,
+            "Prepared object timeout was not delivered near its deadline"
+        );
     }
 
     private synchronized void replay(S3ObjectControlManager manager, List<ApiMessageAndVersion> records) {


hachiwar:
Thanks @BackendArchitectX for tracing the timer history in #1603 and obtaining the intended-policy confirmation from @superhx. This PR implements that confirmed behavior and adds the timer-path regression coverage.

superhx (approved):

PR #3545, opened 2026-08-26
fix(config): use quorum voter endpoint for wildcard controller listener

BackendArchitectX:
## Summary

When a controller listener uses a wildcard host and no explicit controller dvertised.listeners is configured, use the current node's endpoint from controller.quorum.voters as the effective advertised controller listener.

This avoids resolving the wildcard listener to a canonical hostname that may not be reachable by other nodes.

Explicit dvertised.listeners remains authoritative.

## Changes

- Parse explicitly configured advertised controller listeners separately from the listener fallback.
- For the first wildcard controller listener, use the matching local node endpoint from controller.quorum.voters.
- Preserve the existing behavior for explicitly advertised and non-wildcard controller listeners.
- Add a regression test verifying both the voter hostname and port are used.

## Testing

- :core:S3UnitTest regression test passed.
- All related KafkaConfigTest.testEffectAdvertiseControllerListener* tests passed.
- 
at passed.
- :core:checkstyleMain passed.
- git diff --check passed.

This behavior is consistent with the controller advertised-listener fallback addressed upstream in Apache Kafka KAFKA-20380.

Related to #2833

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaConfig.scala
+++ b/core/src/main/scala/kafka/server/KafkaConfig.scala
@@ -872,13 +872,35 @@ class KafkaConfig private(doLog: Boolean, val props: util.Map[_, _])
   }
 
   def effectiveAdvertisedControllerListeners: Seq[EndPoint] = {
-    val controllerAdvertisedListeners = advertisedListeners.filter(l => controllerListenerNames.contains(l.listenerName.value()))
+    val controllerAdvertisedListeners = Option(getString(SocketServerConfigs.ADVERTISED_LISTENERS_CONFIG))
+      .map(CoreUtils.listenerListToEndPoints(_, effectiveListenerSecurityProtocolMap, requireDistinctPorts = false))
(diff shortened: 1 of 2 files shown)


superhx:
Thanks for the fix and for adding the regression test.

This behavior is already addressed upstream in Apache Kafka via KAFKA-20380 / apache/kafka#21918, which uses the current node's endpoint from `controller.quorum.voters` when `advertised.listeners` is not explicitly configured.

To keep AutoMQ behavior aligned with Apache Kafka and avoid maintaining a separate implementation, we will pick up this change when the corresponding upstream updates are synchronized into AutoMQ. I’m closing this PR for now, and we can revisit it if the upstream change is not included in the next synchronization.

BackendArchitectX:
Thanks for the clarification. Understood. We’ll rely on the upstream synchronization for KAFKA-20380. I’ll revisit this only if the change is not included afterward.

PR #3546, opened 2026-08-28
fix(zerozone): advertise proxy leader in KIP-951 responses

deniszh:
KIP-951 responses returned the proxy endpoint while keeping the real partition leader ID. Clients such as franz-go trusted the ID and routed requests back to the wrong broker, causing repeated NOT_LEADER_OR_FOLLOWER errors.

Use the routed endpoint ID as the advertised leader and cover the case where proxy and partition leader IDs differ.

### DIFF
--- a/core/src/main/scala/kafka/server/streamaspect/ElasticKafkaApis.scala
+++ b/core/src/main/scala/kafka/server/streamaspect/ElasticKafkaApis.scala
@@ -254,7 +254,8 @@ class ElasticKafkaApis(
           case None => (-1, -1)
         }
     }
-    LeaderNode(leaderId, leaderEpoch, OptionConverters.toScala(trafficInterceptor.getLeaderNode(leaderId, clientIdMetadata, ln.value())))
+    val node = OptionConverters.toScala(trafficInterceptor.getLeaderNode(leaderId, clientIdMetadata, ln.value()))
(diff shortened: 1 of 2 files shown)


superhx (approved):

superhx:
Thanks for tracking this down and fixing it. I did some follow-up validation to clarify the impact of the inconsistent leader ID and ZeroZone endpoint.

With a single-partition topic and a franz-go `client.id` whose rack differed from the partition rack, the old behavior was reproducible:

- **franz-go** trusts `CurrentLeader.LeaderID` and routes to the real broker. A complete KIP-951 hint suppresses the immediate Metadata refresh (`wanted_metadata_update: true, triggering_metadata_update: false`), so it recovers only on another refresh trigger or the periodic refresh. With the default `MetadataMaxAge`, that is approximately **0–5 minutes**, and records may time out before recovery.
- **Kafka Java client** may also apply the incorrect hint temporarily, but Produce/Fetch requests another Metadata update. It normally recovers after the refresh backoff plus one Metadata round trip, typically **a few hundred milliseconds to about one second** with default settings.

So the issue can affect both clients, but the impact is materially higher for franz-go because the incorrect route may persist until the next periodic Metadata refresh. The fix correctly makes the advertised leader ID consistent with the proxy endpoint.

PR #3548, opened 2026-08-29
fix(zerozone): advertise proxy leader in KIP-951 responses (#3546)

superhx:
KIP-951 responses returned the proxy endpoint while keeping the real partition leader ID. Clients such as franz-go trusted the ID and routed requests back to the wrong broker, causing repeated NOT_LEADER_OR_FOLLOWER errors.

Use the routed endpoint ID as the advertised leader and cover the case where proxy and partition leader IDs differ.


superhx:
cherry-pick https://github.com/AutoMQ/automq/pull/3546

1sonofqiu (approved):

PR #3554, opened 2026-08-31
fix(s3stream): retry block index load after compaction race

### LINKED ISSUE #3550: Block index load retry reuses a failed future after object compaction
# Block index load retry reuses a failed future after object compaction

## Summary

`StreamReader` treats `BlockNotContinuousException` as recoverable, resets its
cached blocks, and retries the read. However, the failed block-index load
remains referenced by `inflightLoadIndexCf`. The retry therefore reuses the
already-failed future instead of calling `objectManager.getObjects()` again.

A transient object-compaction race can consequently fail the current read even
after the metadata image has converged. For Kafka log cleaning, the effect is
more persistent: Kafka marks the partition as uncleanable and stops compacting
it until the broker-local in-memory state is cleared.

## Example failure flow

Assume a stream initially has three source objects:

```text
Object A: [0, 100)
Object B: [100, 200)
Object C: [200, 300)
```

A reader has already loaded blocks through offset `100` and needs to continue
reading `[100, 300)`. At the same time, object compaction replaces `B` and `C`
with a new object:

```text
Before compaction:
  Object B: [100, 200)
  Object C: [200, 300)

After compaction:
  Object D: [100, 300)
```

The compaction result is valid once its metadata update has converged. The
reader can nevertheless combine blocks loaded by two calls made against
different metadata images. For example, an earlier call can cache blocks from
the old object:

```text
old image, Object B: [100, 150), [150, 200)
```

When the reader later extends its window from offset `200`, a new
`getObjects()` call can see only compacted `Object D`. The block containing
offset `200` can start before that offset:

```text
new image, Object D: [100, 180), [180, 240), [240, 300)
```

`putBlock()` has accepted the cached blocks from `B`, leaving
`lastBlock.endOffset=200`. `ObjectReader.find(..., 200, ...)` can return the
`[180, 240)` block from `D`, because that block contains the requested offset.
The continuity invariant then fails:

```text
expected next start offset: 200
actual next start offset:   180
```

This overlap causes `BlockNotContinuousException`. A gap produces the same
result; for example, an old block ending at `200` followed by a new-view block
starting at `220` also fails the continuity check.

The reader is designed to recover from this transient view:

1. `read(..., leftRetries)` catches `BlockNotContinuousException`.
2. `resetBlocks()` discards the mixed block list.
3. The retry should call `getObjects()` again.
4. With the converged image, the retry should see only `Object D`, load
   `[100, 180)`, `[180, 240)`, `[240, 300)`, and complete the original read.

The bug changes steps 3 and 4. The first load has completed
`inflightLoadIndexCf` exceptionally but has not cleared the field. The retry
therefore finds a non-null future and chains from that same failure:

```text
first attempt
  -> reader retains old-view blocks from Object B through offset 200
  -> next getObjects() sees compacted Object D
  -> block overlap at offset 200/180
  -> BlockNotContinuousException
  -> inflightLoadIndexCf completes exceptionally and remains assigned

retry
  -> resetBlocks()
  -> sees inflightLoadIndexCf != null
  -> reuses the already-failed future
  -> does not call getObjects() again
  -> cannot observe the converged D-only view
  -> original read fails
```

The failure then propagates through `StreamReaders`, `S3Storage`, and
`S3Stream`.

```text
object compaction
  -> transient mixed object view
  -> non-continuous block indexes
  -> BlockNotContinuousException
  -> resetBlocks()
  -> retry reuses failed inflightLoadIndexCf
  -> no fresh metadata load
  -> current read fails
```

## Expected behavior

After `BlockNotContinuousException`, the retry should start a new block-index
load. Once the compaction metadata image has converged, the new
`objectManager.getObjects()` result should be continuous and the same read
should succeed.

## Actual behavior

The failed `inflightLoadIndexCf` is cleared only on the successful completion
path. The retry inherits the first attempt's failure and cannot observe the
new metadata image.

### `__consumer_offsets` and other compacted partitions during Kafka log cleaning

If Kafka's log cleaner encounters this failure while compacting a partition,
it catches the `LogCleaningException` and adds that partition to
`LogCleanerManager.uncleanablePartitions`.

Kafka excludes the partition from subsequent cleaning attempts. For
`__consumer_offsets`, compaction of the affected partition therefore stops and
stays stopped even though the underlying object-compaction race was transient.
The uncleanable marker is broker-local in-memory state and has no public Kafka
CLI operation that clears it directly.

A partition reassignment can be used as a temporary operational workaround. It
removes and recreates the broker-local partition log state, allowing the
partition to become eligible for log cleaning again. This clears the symptom,
but does not fix the block-index retry bug; a later race can mark the partition
uncleanable again.

### Normal topic reads

If a normal Kafka fetch encounters the same race, the current fetch can fail.
`StreamReaders` closes and does not retain the failed `StreamReader`. A later
fetch creates or uses a fresh reader and should succeed after the metadata
image has converged. The impact is therefore normally limited to the request
that overlaps the race, rather than permanently making the partition
unreadable.

This distinction matters: Kafka log cleaning converts one read failure into a
persistent compaction stop, while a normal consumer can retry the failed fetch.

## Relevant code

- `s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java`
  - `read(..., leftRetries)` classifies `BlockNotContinuousException` as
    recoverable and invokes reset/retry.
  - `loadMoreBlocksWithoutData0()` stores the shared in-flight future and does
    not clear it on failure.
  - `putBlock()` enforces block-index continuity.
- `s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReaders.java`
  - closes the failed reader and records `from block cache fail`.
- `s3stream/src/main/java/com/automq/stream/s3/S3Storage.java`
  - propagates the block-cache failure to the stream read.
- `core/src/main/scala/kafka/log/LogCleaner.scala`
  - marks a partition uncleanable after an unexpected cleaning exception.
- `core/src/main/scala/kafka/log/LogCleanerManager.scala`
  - excludes entries in `uncleanablePartitions` from later cleaning attempts.

## Proposed fix

Detach and clear `inflightLoadIndexCf` before completing it on both success and
failure. Use a local reference so the completion callback resolves the future
owned by that load:

```java
findBlockIndexesCf.whenCompleteAsync((nil, ex) -> {
    CompletableFuture<Void> cf = inflightLoadIndexCf;
    inflightLoadIndexCf = null;
    if (ex != null) {
        cf.completeExceptionally(ex);
        return;
    }
    GET_INDICES_TIME_FIND_INDEX_STATS.record(time.elapsedAs(TimeUnit.NANOSECONDS));
    cf.complete(null);
}, eventLoop);
```

## Regression coverage

Add a test with the following sequence:

1. The first `getObjects()`/block-index load returns a non-continuous sequence.
2. The first attempt throws `BlockNotContinuousException`.
3. The reader resets its block state.
4. The retry calls `getObjects()` again instead of reusing the failed future.
5. The second load returns a continuous sequence.
6. The original `read()` completes successfully.

The test should also assert that `inflightLoadIndexCf` no longer retains an
exceptionally completed future.

BackendArchitectX:
## Why

Forward-port the fix from #3551 to main.

After object compaction, StreamReader can temporarily combine block indexes from different metadata images and raise BlockNotContinuousException.

The exception is recoverable, but the failed inflightLoadIndexCf remains assigned. The retry therefore reuses the exceptionally completed future instead of loading the converged metadata image.

Fixes #3550.

## What

Allow a recoverable block-index load failure to start a fresh metadata load during the same read.

This forward-ports #3551 to main together with its regression coverage.

## How

Each block-index load owns a local CompletableFuture.

The completion callback clears the shared inflightLoadIndexCf slot before completing the load-owned future on both success and failure. This prevents the retry from observing and reusing an already-completed failed future.

## Verification

- ./gradlew :s3stream:test --tests com.automq.stream.s3.cache.blockcache.StreamReaderTest.testReadRetryAfterDiscontinuousBlockIndexLoad
- ./gradlew :s3stream:test --tests com.automq.stream.s3.cache.blockcache.StreamReaderTest
- ./gradlew :s3stream:checkstyleMain :s3stream:checkstyleTest :s3stream:spotlessJavaCheck
- git diff HEAD^ HEAD --check

All passed locally.

## Reviewer notes

This is a direct forward-port of the reviewed and merged 1.8 fix in #3551 with no main-specific changes.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
@@ -426,7 +426,8 @@ import static com.automq.stream.utils.FutureUtil.exec;
             return CompletableFuture.completedFuture(null);
         }
         long currentBlocksEpoch = blocksEpoch;
-        inflightLoadIndexCf = new CompletableFuture<>();
+        CompletableFuture<Void> loadIndexCf = new CompletableFuture<>();
+        inflightLoadIndexCf = loadIndexCf;
         long nextLoadingOffset = calWindowBlocksEndOffset();
         AtomicLong nextFindStartOffset = new AtomicLong(nextLoadingOffset);
         TimerUtil time = new TimerUtil();
@@ -467,16 +468,15 @@ import static com.automq.stream.utils.FutureUtil.exec;
             return prevCf;
         }, eventLoop);
         findBlockIndexesCf.whenCompleteAsync((nil, ex) -> {
+            inflightLoadIndexCf = null;
             if (ex != null) {
-                inflightLoadIndexCf.completeExceptionally(ex);
+                loadIndexCf.completeExceptionally(ex);
                 return;
             }
             GET_INDICES_TIME_FIND_INDEX_STATS.record(time.elapsedAs(TimeUnit.NANOSECONDS));
-            CompletableFuture<Void> cf = inflightLoadIndexCf;
-            inflightLoadIndexCf = null;
-            cf.complete(null);
+            loadIndexCf.complete(null);
         }, eventLoop);
-        return inflightLoadIndexCf;
+        return loadIndexCf;
     }
 
     private static DeltaHistogram readBlockCacheStats(boolean isCacheHit) {
--- a/s3stream/src/test/java/com/automq/stream/s3/cache/blockcache/StreamReaderTest.java
+++ b/s3stream/src/test/java/com/automq/stream/s3/cache/blockcache/StreamReaderTest.java
@@ -36,6 +36,7 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.mockito.InOrder;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -58,6 +59,7 @@ import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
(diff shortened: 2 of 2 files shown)


superhx:
This is duplicated to https://github.com/AutoMQ/automq/pull/3553

PR #3555, opened 2026-08-31
fix(s3stream): prevent closed reader from restoring blocks

BackendArchitectX:
## Why

StreamReader.close() clears the current block state, but block-index/readahead work can still continue around the close lifecycle boundary.

There are two cases:

1. An in-flight block-index/readahead load can complete after close().
2. Queued or follow-up readahead work can start a new block-index load after close().

For work that started before close(), the load can still see the same blocksEpoch and add blocks back into a reader that has already been closed.

For work that starts after close(), epoch invalidation alone is not sufficient because the new load observes the updated epoch.

## What

Prevent both in-flight and post-close block-index loads from restoring block state after a StreamReader is closed.

## How

Increment blocksEpoch in close() before clearing the current blocks.

Existing block-index loading already checks the captured epoch before continuing and before inserting blocks, so changing the epoch prevents work started before close() from restoring stale state.

Additionally, prevent loadMoreBlocksWithoutData0() from starting a new block-index load once the StreamReader is closed. This covers queued readahead and follow-up load continuations that begin after close().

Together, these checks ensure that a closed StreamReader cannot repopulate its block state through either existing in-flight work or newly started readahead work.

## Verification

Added regression coverage for both cases:

- testClosePreventsInflightReadaheadFromRestoringBlocks
  - Starts readahead with a delayed metadata load.
  - Closes the reader before the load completes.
  - Completes the delayed load and verifies that blocksMap remains empty.

- testClosedReaderDoesNotStartNewReadaheadLoad
  - Closes the reader before starting readahead.
  - Attempts readahead after close().
  - Verifies that ObjectManager#getObjects is not invoked and blocksMap remains empty.

Local verification:

- ./gradlew :s3stream:test --tests com.automq.stream.s3.cache.blockcache.StreamReaderTest
- ./gradlew :s3stream:checkstyleMain :s3stream:checkstyleTest :s3stream:spotlessJavaCheck
- git diff --check

All passed locally.

### DIFF
--- a/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
+++ b/s3stream/src/main/java/com/automq/stream/s3/cache/blockcache/StreamReader.java
@@ -186,6 +186,7 @@ import static com.automq.stream.utils.FutureUtil.exec;
 
     public void close() {
         closed = true;
+        blocksEpoch++;
         List<Block> blocks = new ArrayList<>(blocksMap.values());
         // The Block#markRead will immediately invoke after the Block is removed.
         blocksMap.clear();
@@ -421,6 +422,9 @@ import static com.automq.stream.utils.FutureUtil.exec;
     }
 
     private CompletableFuture<Void> loadMoreBlocksWithoutData0(long endOffset) {
+        if (closed) {
+            return CompletableFuture.completedFuture(null);
+        }
         if (inflightLoadIndexCf != null) {
             return inflightLoadIndexCf.thenCompose(rst -> loadMoreBlocksWithoutData0(endOffset));
         }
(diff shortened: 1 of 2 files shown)


BackendArchitectX:
Hi @Gezi-lzq — could you please take a look at #3555 when you have a chance?

The implementation changed materially after the original Aug 31 review request. The Sep 8 version added the `closed` guard needed to prevent queued/follow-up block-index loads after `StreamReader.close()`, in addition to epoch invalidation for already in-flight work.

I've now linked bug #3615 and validated both regressions against current `main` (`7d4f733348`): each fails without the production fix, while both and the complete `StreamReaderTest` pass with #3555 applied. Style and diff checks also pass.

The production change remains limited to four added lines.

Thanks.

Gezi-lzq:
Thanks for the heads-up! I'll take a look and review it as soon as I can.

PR #3579, opened 2026-09-04
feat(metrics): support Basic authentication for Prometheus endpoint

### LINKED ISSUE #3573: [Enhancement] Support authentication for the built-in Prometheus metrics endpoint
### Who is this for and what problem do they have today?

AutoMQ operators running either VM/IaaS or Kubernetes deployments who expose the built-in Prometheus pull endpoint to a remote Prometheus-compatible collector.

Today, `s3.telemetry.metrics.exporter.uri` only configures the Prometheus listener host and port, for example:

```properties
s3.telemetry.metrics.exporter.uri=prometheus://?host=0.0.0.0&port=9090
```

The current implementation starts the endpoint without authentication:

- `automq-metrics/src/main/java/com/automq/opentelemetry/exporter/PrometheusMetricsExporter.java`
- `automq-metrics/src/main/java/com/automq/opentelemetry/exporter/MetricsExporterURI.java`

Consequently, any client with network access to the listener can scrape `/metrics`. Operators must currently rely on network isolation, a local collector, or an externally managed nginx/Envoy proxy.

### Why is solving this problem impactful?

Metrics may expose operational metadata and detailed cluster behavior. Deployments with security or compliance requirements need a supported way to prevent anonymous scraping without inventing a different integration for each environment.

The solution should cover both deployment models:

- VM/IaaS: a local collector, a host-level proxy, or an authenticated in-process endpoint.
- Kubernetes: a node/pod-local collector, a sidecar proxy, or an authenticated in-process endpoint.

### Additional notes

This issue intentionally does not prescribe the final architecture. The design should evaluate at least these options:

1. In-process authentication using `com.sun.net.httpserver.Authenticator`.
2. A deployment-level proxy that provides HTTPS, Basic Auth, or mTLS.
3. Keeping the endpoint loopback-only and using a local collector that pushes metrics to the backend.

Relevant implementation details found during the initial investigation:

- AutoMQ currently uses `opentelemetry-exporter-prometheus:1.40.0-alpha`. Its `PrometheusHttpServerBuilder` does not expose `setAuthenticator(...)`.
- The underlying Prometheus Java `HTTPServer.Builder` already exposes `authenticator(...)` and `httpsConfigurator(...)`.
- OpenTelemetry `PrometheusHttpServerBuilder` exposes `setAuthenticator(...)` starting with `1.50.0-alpha`. Adopting it would require an aligned OpenTelemetry dependency upgrade and compatibility testing.
- An `Authenticator` alone does not encrypt traffic. Basic credentials over plain HTTP are insufficient for untrusted networks; HTTPS/mTLS requirements must be decided separately.
- Credentials must not be placed directly in the exporter URI because the URI is logged. Prefer a mounted secret file or an existing secret-provider abstraction, with explicit rotation/reload semantics.

Decision points:

- Supported authentication modes: Basic, Bearer token, mTLS, or a subset.
- Whether TLS is implemented in-process or delegated to a proxy.
- Configuration contract and backward compatibility. Authentication should remain opt-in for existing deployments.
- Credential provisioning and rotation for both VM/IaaS and Kubernetes.
- Whether `/`, `/metrics`, and `/-/healthy` share the same authentication policy.
- Collector configuration and migration from existing unauthenticated port `9090` targets.

Expected acceptance outcomes:

- A documented, supported path exists for both VM/IaaS and Kubernetes deployments.
- An unauthenticated remote client cannot read metrics when authentication is enabled.
- An authorized Prometheus-compatible collector can scrape metrics successfully.
- Secrets are not emitted in logs or stored as plaintext in the exporter URI.
- Existing deployments remain compatible when authentication is not enabled.
- Automated tests cover successful authentication, rejected credentials, endpoint behavior, and shutdown lifecycle.

BackendArchitectX:
Fixes #3573

## Why
The built-in Prometheus endpoint currently allows any client with network access to scrape /metrics. This change adds optional in-process HTTP Basic authentication while preserving the existing unauthenticated behavior by default.

## What
- Add optional Basic authentication for /metrics and /metrics/...
- Keep authentication disabled by default for backward compatibility
- Keep / and /-/healthy public
- Add broker configs for auth type, username, and password
- Keep credentials outside s3.telemetry.metrics.exporter.uri
- Register the password as a Kafka PASSWORD config
- Propagate an optional authenticator through the telemetry path
- Upgrade OpenTelemetry to versions exposing PrometheusHttpServerBuilder#setAuthenticator
- Preserve existing Prometheus behavior with MemoryMode.IMMUTABLE_DATA
- Add authentication, configuration, endpoint-policy, and lifecycle tests

## Configuration
s3.telemetry.metrics.prometheus.auth.type=none|basic
s3.telemetry.metrics.prometheus.auth.username=<username>
s3.telemetry.metrics.prometheus.auth.password=<password>

The default is none, so existing deployments remain unchanged. When basic is enabled, missing or invalid credentials return HTTP 401 and valid credentials can scrape metrics successfully.

## OpenTelemetry compatibility
Upgrades OpenTelemetry SDK to 1.50.0, SDK/Prometheus alpha to 1.50.0-alpha, and instrumentation alpha to 2.16.0-alpha. Compatibility updates preserve existing metric cardinality, local JMX, GC metric shape, and Prometheus immutable-data behavior.

## Security considerations
Basic authentication does not encrypt transport. Deployments crossing untrusted networks should continue to provide TLS through an ingress, reverse proxy, sidecar, or equivalent network control. Credential rotation is restart-based in this initial implementation.

## Testing
Validated locally with :automq-metrics:test, :automq-metrics:check, AutoMQConfigTest, :core:checkstyleMain, :core:spotlessJavaCheck, jar -x test, rat, and git diff --check upstream/main...HEAD.

Coverage includes default unauthenticated scraping, missing credentials, wrong username/password, malformed Base64, wrong auth scheme, successful authenticated scraping, /metrics/... protection, public root/health endpoints, config validation, and exporter shutdown/port release.


BackendArchitectX:
@1sonofqiu @Gezi-lzq  @superhx, could you please take a look at the proposed direction before I extend this further?

#3573 intentionally leaves the authentication architecture open. #3579 currently takes the in-process Basic Auth approach with:

authentication opt-in and disabled by default
credentials outside the exporter URI
/metrics protected while / and /-/healthy remain public
TLS delegated to the deployment layer
restart-based credential rotation
the OpenTelemetry upgrade required for setAuthenticator()

Before I make further changes, could you confirm that the in-process approach and the /metrics-only authentication policy are acceptable?

If this direction is accepted, I can then tighten the configuration validation, add broker-level/CI-backed coverage and complete the deployment documentation.

1sonofqiu:
@BackendArchitectX Thank you for your contribution ! I will review it in the next few days.

PR #3582, opened 2026-09-06
chore(clients): suppress unavoidable X509Certificate DN deprecation warnings

brittytino:
## Why

Fixes #3282.

`CommonNameLoggingTrustManagerFactoryWrapper.NeverExpiringX509Certificate` extends `X509Certificate` directly. `getSubjectDN()`/`getIssuerDN()` are `abstract` on that class (deprecated since JDK 16, but never made non-abstract or removed), so any direct subclass is still required to implement them. The build currently emits 4 deprecation warnings from this file (one at each override declaration, one at each internal delegating call).

## What / How

I looked at swapping the internal delegation to `getSubjectX500Principal()`/`getIssuerX500Principal()` as suggested in the issue, but that changes the *runtime type* of the returned `Principal`: the original certificate's `getSubjectDN()` typically returns an implementation-specific `Principal` (e.g. `sun.security.x509.X500Name`), while `getSubjectX500Principal()` returns a `javax.security.auth.x500.X500Principal`. I verified with a real self-signed cert that these two are **not** `.equals()` to each other even though their `toString()` output is identical:

```
subjectDN class: class sun.security.x509.X500Name
subjectX500 class: class javax.security.auth.x500.X500Principal
subjectDN.equals(subjectX500): false
```

Making that swap would silently change the wrapper's observable behavior and would break the existing test (`CommonNameLoggingTrustManagerFactoryWrapperTest#testNeverExpiringX509Certificate`), which asserts `cert.getIssuerDN()` / `cert.getSubjectDN()` equal `wrappedCert.getIssuerDN()` / `wrappedCert.getSubjectDN()`.

Since the overrides can't be removed and changing their delegation isn't behavior-preserving, this PR keeps the delegation exactly as-is and adds `@SuppressWarnings("deprecation")` on the two required overrides (with a short comment explaining why they can't be removed), plus on the test method that intentionally exercises the deprecated methods to verify backward-compatible behavior. No functional change.

## Testing

- `./gradlew :clients:compileJava :clients:compileTestJava` — the 4 warnings from this file are gone; only pre-existing, unrelated warnings remain (`AccessController`/`Subject.getSubject`, `JsonNode#fields()`, `ObjectMapper#setSerializationInclusion`).
- `./gradlew :clients:test --tests "org.apache.kafka.common.security.ssl.CommonNameLoggingTrustManagerFactoryWrapperTest"` — all 11 tests pass.
- `./gradlew --build-cache :clients:checkstyleMain :clients:checkstyleTest :clients:spotbugsMain :clients:spotbugsTest :clients:spotlessJavaCheck` — all pass, no formatting changes.


Gezi-lzq:
Thanks for investigating this carefully. There are already multiple pull requests addressing #3282, and we don’t want to continue parallel implementations for the same cleanup. We’ll consolidate the decision in the issue and close this duplicate PR.

PR #3592, opened 2026-09-16
fix(metadata): remove empty partition stream mappings

superhx:
## Why

Deleting the last stream for a topic partition leaves an empty set in `partition2streams`. The latest metadata image continues to retain the partition key and set even after old snapshots are released. Repeated topic deletion and recreation with new topic UUIDs therefore accumulates stale entries; health-check topic churn is one possible trigger.

## What

Remove touched partition mappings when their final stream set is empty. Preserve non-empty mappings, same-delta stream replacement, and historical snapshot contents. No protocol, configuration, or serialized metadata changes.

## How

After applying stream deletions and additions under the existing registry lock, inspect only partition sets touched by this delta and remove empty entries through `TimelineHashMap`. Deferring cleanup until additions complete prevents a replacement stream from being added to a cached set whose mapping has already been removed. Existing copy-on-write updates preserve historical snapshot values.

Regression tests cover repeated creation/deletion with distinct topic UUIDs, forward/reverse map cleanup, historical snapshot retention, and deleting/replacing the last stream in one delta.

## Reviewer notes

Review the partition mapping update in `S3StreamsMetadataDelta.apply()` and the regression tests in `S3StreamsMetadataImageTest`. This is a prospective cleanup of touched partitions, not a full scan to purge existing stale entries. The CMP health-check recreation logic is outside this PR's scope.


1sonofqiu (approved):

superhx:
Local verification passed with Java 17: `./gradlew --build-cache metadata:S3UnitTest rat checkstyleMain checkstyleTest spotlessJavaCheck` (BUILD SUCCESSFUL). The metadata S3Unit suite contains 149 tests with zero failures. The new cleanup regression was confirmed failing on the original implementation before applying the fix. `rat` was skipped because the repository sets `rat.enabled = false`. Independent read-only review found no correctness issues. Cleanup is deferred until stream additions complete to preserve same-delta replacement through the per-delta set cache.

PR #3599, opened 2026-09-17
fix(s3stream): complete lazy commits when the WAL cache is empty

superhx:
## Why

PR #3513 bound lazy commits only when a new WAL upload context was created. A global commit with an empty active WAL cache therefore remained queued indefinitely, even when there were no inflight uploads. ZeroZone waits for this commit before reporting its node committed epoch, so an idle broker could prevent the global committed epoch from advancing.

## What

Complete global lazy commit barriers even when no active WAL block is archived. Preserve stream-scoped force-upload behavior and the existing upload return value.

## How

In the no-block branch, drain queued lazy commits only for `MATCH_ALL_STREAMS` and bind their commit and trim futures to a snapshot of existing inflight uploads. An empty snapshot completes immediately; otherwise the barriers wait for those uploads. Stream-scoped attempts without matching active data leave lazy commits queued.

## Reviewer notes

Read `S3Storage.uploadDeltaWAL(long, boolean)` alongside `notifyLazyUpload`. `S3StorageTest.testEmptyWALCompletesGlobalLazyCommit` covers successful completion of the trim-awaiting commit on an empty WAL. The existing stream-scoped regression test protects the unchanged boundary.


superhx:
Local verification: `git diff --check` and `./gradlew --build-cache s3stream:test rat checkstyleMain checkstyleTest spotlessJavaCheck` passed. The build configuration skipped `rat`. The initial full-suite run failed in `AsyncNetworkBandwidthLimiterTest.testThrottleConsumeWithLargeChunk` (expected 0, actual 1048576); the subsequent full-suite run passed, including this test and the new empty-WAL regression test. An independent static review of the scoped diff found no actionable issues.

Gezi-lzq (approved):

PR #3609, opened 2026-09-20
fix(log): respect log start offset in timestamp lookup

shantanukawale-stripe:
## Why

AutoMQ can advance a partition's log start offset before the corresponding stream segments are physically removed. During that interval, timestamp-based ListOffsets searches could select a segment that was fully below the new log start offset. The elastic segment lookup also began scanning at the time-index offset without honoring the caller's starting offset.

Together, those behaviors allowed a timestamp lookup at a retention boundary to return an expired offset below the LSO or no result even when a later segment contained a valid record.

## What

Timestamp searches now begin at the segment whose base-offset range contains the requested start offset, excluding all earlier segments without reading their tails. Elastic segment scans never begin before that offset. The search also continues through later eligible segments if a partially retained segment's historical maximum timestamp does not yield a valid record after the LSO.

There are no wire-protocol, configuration, or storage-format changes.

## How

`UnifiedLog.searchOffsetInLocalLog` snapshots the segment collection as before, then:

1. uses in-memory segment base offsets to skip segments before the range containing the requested start offset, without calling `readNextOffset` on each segment;
2. keeps the existing largest-timestamp eligibility check;
3. searches eligible segments in order until one returns a valid timestamp/offset pair.

`ElasticLogSegment.findOffsetByTimestamp` now scans from `max(startingOffset, timeIndexOffset)`, matching the logical contract of Kafka's file-backed `LogSegment` while retaining AutoMQ's offset-addressed stream reads.

Focused S3Unit coverage protects both layers:

- a sealed elastic segment whose time index points below `startingOffset`;
- two sealed segments with an LSO inside the first segment and deliberately non-monotonic timestamps, queried at the boundary timestamp and one millisecond later.

The latter asserts that both results are present, monotonic, and at or above the LSO while forcing the search to continue to the later segment.

## Reviewer notes

Suggested reading order:

1. `core/src/main/scala/kafka/log/UnifiedLog.scala` for cross-segment eligibility and continuation.
2. `core/src/main/scala/kafka/log/streamaspect/ElasticLogSegment.java` for the per-segment lower bound.
3. `ElasticLogSegmentTest` and `ElasticUnifiedLogTest` for the sealed-segment and retention-boundary contracts.

The AutoMQ-owned change inside Kafka's `UnifiedLog` is kept within narrow inject markers for future upstream merges.


shantanukawale-stripe:
Local verification on Corretto 17:

- PASS: `./gradlew --build-cache core:S3UnitTest --tests 'kafka.log.streamaspect.ElasticLogSegmentTest.testFindOffsetByTimestampRespectsStartingOffsetInSealedSegment' --tests 'kafka.log.streamaspect.ElasticUnifiedLogTest.testFetchOffsetByTimestampAfterLogStartOffsetTrim'`
- PASS: `./gradlew --build-cache metadata:S3UnitTest core:S3UnitTest s3stream:test`
- PASS: `./gradlew --build-cache rat checkstyleMain checkstyleTest spotlessJavaCheck` (`rat` is skipped by the current Gradle configuration; all remaining requested gates ran successfully.)

shantanukawale-stripe:
Updated the candidate selection to avoid calling `readNextOffset` on every segment. It now uses the in-memory segment base offsets to start at the range containing `startOffset`, then continues across later timestamp-eligible segments. This avoids turning file-backed ListOffsets into per-segment tail reads. The test wording now also matches the boundary / boundary-plus-one-millisecond inputs.

Verification for the updated revision:

- PASS: both focused retention-boundary S3Unit tests
- PASS: `rat checkstyleMain checkstyleTest spotlessJavaCheck` (`rat` is skipped by the current Gradle configuration)
- The full `metadata:S3UnitTest core:S3UnitTest s3stream:test` run completed the changed tests successfully but failed in the unrelated `PartitionSnapshotsManagerTest.testPartitionOpenWakesPendingLongPoll` (`expected 1 but was 0`). An isolated rerun after restarting the Gradle daemon reproduced that same unrelated failure. This PR changes only log timestamp lookup code and its two test classes.

shantanukawale-stripe:
Hosted CI is now fully green: Checkstyle, SpotBugs, Unit Test, aggregate build result, PR title, changes, and CLA all passed. The local PartitionSnapshotsManagerTest result noted above did not reproduce in the hosted Unit Test job.

superhx (approved):

